PR #46, opened 2014-08-13
USERGRID-82 Cannot change custom entity name

stliu:
also, update log4j.xml to avoid duplicate log message
update spring configuration to configure static attribute


GERey:
@stliu If you could address the conflicts I'd be happy to take another look at it! Thanks!

mdunker:
master is now the 2.1+ branch -- this PR therefore is no longer valid. Please submit against 1.x branch if this is still an issue. Thanks!

PR #112, opened 2014-12-01
No source in es

snoopdave:
DO NOT MERGE. I'm creating this for review purposes.

Changes to prevent us from storing source document or source fields in ElasticSearch
See also: https://issues.apache.org/jira/browse/USERGRID-255


tnine:
We'll want to put this in with Shawn's changes on this PR

#161

snoopdave:
I'm going to close this. It is obsolete because of the many changes in two-dot-o-dev. 

Also, conventional wisdom seems to be that having source in ES is very valuable for debugging problems in production UG2 systems.

PR #126, opened 2014-12-19
[USERGRID-288]Updated Geo tests to work with the new test framework

r3b:
https://issues.apache.org/jira/browse/USERGRID-288


r3b:
This was bundled in a later PR.

PR #131, opened 2014-12-23
Fixed PagingEntities and PagingResourceTest.

GERey:
Fixed bug where proper resource wasn't being used in the delete call.
Added a queryParameter get call to the Entity method for testing purposes.
Added a way to get the response from Collection


rodsimpson:
Merged.  Thanks!!

PR #134, opened 2014-12-29
Portal: Use same host and port for override url (default)

deshbirc:
Try and use the same hostname and port for the API overrideUrl. Assumes that Portal and Usergrid are deployed on the same tomcat instance. For example, this would be useful when a single Tomcat instance is being used to host both portal and usergrid apps.


rodsimpson:
Hello, I was not able to fetch this from your repo.  Has it been made private?

PR #136, opened 2015-01-04
Added the Perl logo to the SDKs and Modules section

aweeraman:
Change to include the Perl logo in the SDKs and Modules panel of the portal.


rodsimpson:
Thanks!! This has been merged.

PR #139, opened 2015-01-05
Usergrid 303

r3b:
Updating query tests to work with new test framework


GERey:
Looks good to me! Merged

PR #142, opened 2015-01-06
Usergrid 278 org management tests fix

GERey:
Handles fixing the organization tests and moved them to using the new rest test framework.


GERey:
https://issues.apache.org/jira/browse/USERGRID-278

rodsimpson:
merged.  thanks!

PR #145, opened 2015-01-08
Portal fixes

johnament:
Fixes for two issues.


johnament:
As mentioned on the list, I'm not having issues with api_url.  The change in here is not important and if you want I can reverse the commit.  My IDE was complaining about it though, so maybe it should be quoted (due to _) ?

PR #153, opened 2015-01-21
Usergrid 323 windows phone notifications support

shawnfeldman:


stack/services/src/main/java/org/apache/usergrid/services/notifications/ProviderAdapterFactory.java
```diff
@@ -37,6 +38,7 @@ public static ProviderAdapter getProviderAdapter(Notifier notifier, EntityManage
        switch(notifier.getProvider().toLowerCase()){
            case "apple" : adapter = new APNsAdapter(entityManager,notifier); break;
            case "google" : adapter = new GCMAdapter(entityManager ,notifier); break;
+           case "windows" : adapter = new WNSAdapter(entityManager ,notifier); break;
            case "noop" : adapter = new TestAdapter(notifier); break;
```
tnine: Seems silly, but I think our default case should be to throw an exception if it doesn't match any of the other cases.  That way we get some explicit feedback something was entered incorrectly into the payload.

stack/services/src/main/java/org/apache/usergrid/services/notifications/wns/WNSAdapter.java
```diff
@@ -0,0 +1,162 @@
...
+                        break;
+                    case "badge":
+                        WnsBadge badge;
+                        if (translatedNotification.getMessage() instanceof Integer) {
+                            badge = new WnsBadgeBuilder().value((Integer) translatedNotification.getMessage()).build();
+                        } else {
+                            badge = new WnsBadgeBuilder().value(translatedNotification.getMessage().toString()).build();
+                        }
+                        service.pushBadge(providerId, badge);
+                        break;
```
tnine: Same thing here, do we want to throw/log an exception for when it's not a type we handle?  That would be helpful operationally if users are having problems and can't send.

stack/services/src/main/java/org/apache/usergrid/services/notifications/ProviderAdapterFactory.java
shawnfeldman: fixed

stack/services/src/main/java/org/apache/usergrid/services/notifications/wns/WNSAdapter.java
shawnfeldman: fixed

tnine:
Looks good.

PR #155, opened 2015-01-27
remove lexer from ignore

shawnfeldman:


snoopdave:
Do the generated lexer files really belong in source code control? I'd rather see the lexer classes generated into a temporary directory like target/generated-classes. They are just build artifacts not really our source code.

shawnfeldman:
removed too many line
the core ones aren't in two-0 and are confusing when you switch branches

snoopdave:
Gotcha. Looks good.

PR #156, opened 2015-01-28
fix authentication workflow

shawnfeldman:


sdks/dotnet/samples/notifications/Usergrid.Notifications/Client/PushClient.cs
```diff
@@ -114,7 +116,7 @@ public PushClient(IUsergridHttpClient usergrid, string notifier)
        
         private async Task<JToken> GetDevice(Guid deviceId)
         {
-            var jsonResponse = await usergrid.SendAsync(HttpMethod.Get, "devices/" + deviceId, null);
+            var jsonResponse = await usergrid.SendAsync(HttpMethod.Get, "users/"+userId+"/devices/" + deviceId, null);
```
r3b: why not use String.Format here? It's not consistent with previous methods.

PR #160, opened 2015-02-16
Usergrid 273

tnine:
Fixes several bugs and upgrades our test infrastructure.  Some key changes.  

1) Indexes are deleted after tests
2) Before a JVM starts, Cassandra and ES are changed
3) Aquillian is now used to test the REST tier (not yet complete)


stack/core/src/main/java/org/apache/usergrid/corepersistence/CpSetup.java
```diff
@@ -61,151 +59,106 @@
...
 
         logger.info( "Initialize system keyspace" );
 
-        cass.createColumnFamily( SYSTEM_KEYSPACE, createColumnFamilyDefinition( 
-                SYSTEM_KEYSPACE, APPLICATIONS_CF, ComparatorType.BYTESTYPE ) );
+        migrate();
 
-        cass.createColumnFamily( SYSTEM_KEYSPACE, createColumnFamilyDefinition( 
-                SYSTEM_KEYSPACE, PROPERTIES_CF, ComparatorType.BYTESTYPE ) );
+        cass.createColumnFamily( SYSTEM_KEYSPACE,
```
GERey: Is there a way to check that we've already created these keyspaces? That way it won't show the exception when we try to recreate already existing keyspaces?

stack/core/src/main/java/org/apache/usergrid/corepersistence/GuiceFactory.java
```diff
@@ -0,0 +1,159 @@
...
+
+
+            ConfigurationManager.loadProperties( cpProps );
+        }
+        catch ( Exception e ) {
+            throw new RuntimeException( "Fatal error loading configuration.", e );
+        }
+
+        //this is seriously fugly, and needs removed
+        injector = Guice.createInjector( new GuiceModule( applicationContext ) );
```
GERey: Could we separate this out? It seems like a method called getObject shouldn't do configuration loading either.

stack/core/src/main/resources/usergrid-core-context.xml
```diff
@@ -69,21 +69,21 @@
...
-<!--	<bean name="lockManager" 
+	<!--  locking for a single node -->
+<!--	<bean name="lockManager"
         class="org.apache.usergrid.locking.singlenode.SingleNodeLockManagerImpl" />-->
-	
+
 	<!--  hector based locks -->
-	<!-- Note that if this is deployed in a production cluster, the RF on the keyspace 
-    MUST be updated to use an odd number for it's replication Factor.  Even numbers can 
+	<!-- Note that if this is deployed in a production cluster, the RF on the keyspace
```
GERey: This should probably be documented in bold somewhere as part of a UG setup.

stack/core/src/main/resources/usergrid-core-context.xml
```diff
@@ -91,7 +91,14 @@
...
-	
+
+
+    <bean id="injector"
+   		class="org.apache.usergrid.corepersistence.GuiceFactory">
+   		<constructor-arg ref="cassandraService" />
+        <constructor-arg ref="properties" />
+   	</bean>
+
 	<bean id="cassandraService"
```
GERey: (sadpanda) inconsistant spacing

stack/core/src/test/java/org/apache/usergrid/mq/MessagesIT.java
```diff
@@ -100,13 +102,16 @@ public void testMessages() throws Exception {
...
-        Map<String, Long> counters = qm.getQueueCounters( "/" );
-        LOG.info( "dumping counters...." + counters );
-        LOG.info( JsonUtils.mapToFormattedJsonString( counters ) );
-        assertEquals( 1, counters.size() );
-        assertNotNull( counters.get( "/foo/bar/" ) );
-        assertEquals( new Long( 3 ), counters.get( "/foo/bar/" ) );
+
+        //wait for counters for flush\
+
+        //TODO Re-evaluate queues and make a cleaner interface
```
GERey: So currently there is no way to get counters from the queue in this branch?

stack/pom.xml
```diff
@@ -1451,375 +1462,80 @@
...
-          <groupId>org.codehaus.mojo</groupId>
-          <artifactId>javancss-maven-plugin</artifactId>
-          <version>2.0</version>
-        </plugin>
+                <plugin>
+                  <groupId>org.apache.maven.plugins</groupId>
+                  <artifactId>maven-compiler-plugin</artifactId>
+                  <version>3.2</version>
+                  <configuration>
+                    <source>1.7</source>
```
GERey: Is this for jdk 1.7? Would this cause incompatibilities in 1.8 when we eventually make the switch?

stack/rest/src/test/java/org/apache/usergrid/rest/AbstractRestIT.java
```diff
@@ -39,23 +39,33 @@
...
+ * Base class for testing Usergrid Jersey-based REST API. Implementations should model the
+ * paths mapped, not the method names. For example, to test the the "password" mapping on
+ * applications.users.UserResource for a PUT method, the test method(s) should following the
  * following naming convention: test_[HTTP verb]_[action mapping]_[ok|fail][_[specific
  * failure condition if multiple]
  */
-//@Concurrent()
+//
+@RunWith( Arquillian.class )
 public abstract class AbstractRestIT extends JerseyTest {
```
GERey: Soon AbstractRestIt will be depreciated once the rest tests have been converted over. Is all we need is the createTestArchieve method to work with arquillian?

stack/core/src/main/resources/usergrid-core-context.xml
tnine: Agreed.  We need to rework our system system as well.

stack/core/src/test/java/org/apache/usergrid/mq/MessagesIT.java
tnine: Yeah, Queue is generally broken. IMHO this feature (queues) should be removed from 2.0 completely since it's unstable at best.  We should focus on entities and graph first, then build our Queue features.

stack/pom.xml
tnine: Nah, just requires the source to be 1.7 compatible.  IF we build it with 1.8, we're still good.

stack/rest/src/test/java/org/apache/usergrid/rest/AbstractRestIT.java
tnine: Yeah, just the @RunwithAttribute and the static method telling it how to create the testable archive.  Everything else is encapsulated within the arquillian.xml file

PR #161, opened 2015-02-18
Usergrid 273 indexbuffer

shawnfeldman:
buffer all indexing calls into sets of 100
bring metrics factory to cp


stack/corepersistence/common/src/main/java/org/apache/usergrid/persistence/core/future/ObservableFuture.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+package org.apache.usergrid.persistence.core.future;
+
+import rx.Observable;
+import rx.Subscriber;
+
+/**
+ *
+ */
+public class ObservableFuture<T> implements Observable.OnSubscribe<T> {
```
tnine: Do we ned this class?  Can't we just use Observable.from(Future<T>)?

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsEntityIndexBatchImpl.java
```diff
@@ -87,41 +86,34 @@
...
         this.client = client;
+        this.indexBatchBuffer = indexBatchBuffer;
         this.failureMonitor = failureMonitor;
         this.entityIndex = entityIndex;
         this.indexIdentifier = IndexingUtils.createIndexIdentifier(config, applicationScope);
         this.alias = indexIdentifier.getAlias();
         this.refresh = config.isForcedRefresh();
-        this.autoFlushSize = autoFlushSize;
-        initBatch();
+        this.promises = new ConcurrentLinkedQueue<>();
```
tnine: Do we want to set an upper bound on this in case we can't flush fast enough?

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsEntityIndexBatchImpl.java
```diff
@@ -228,58 +217,29 @@ public EntityIndexBatch deindex( final IndexScope indexScope, final CandidateRes
 
     @Override
     public void execute() {
-        execute( bulkRequest.setRefresh( refresh ) );
+//        indexBatchBuffer.flush();
+        Observable.from(promises)
```
tnine: See previous comment.  I think you can just do

```
Observable.from(promises).toBlocking().lastOrDefault(null)
```

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/IndexBatchBufferImpl.java
```diff
@@ -0,0 +1,226 @@
...
+                        execute(config.isForcedRefresh());
+                    }
+                });
+        consumer.subscribe();
+    }
+
+    @Override
+    public BetterFuture<ShardReplicationOperationRequestBuilder> put(IndexRequestBuilder builder){
+        RequestBuilderContainer container = new RequestBuilderContainer(builder);
+        metricsFactory.getCounter(IndexBatchBuffer.class,"index.buffer.size").inc();
```
tnine: Can we declare these in the constructor so we don't take the performance hit of the get on every put?

stack/corepersistence/common/src/main/java/org/apache/usergrid/persistence/core/future/ObservableFuture.java
shawnfeldman: will remove, had it as a holdover when i was trying to do pure observables

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/IndexBatchBufferImpl.java
shawnfeldman: we can, i did this because i knew all the keys were cached in a singleton hashmap

shawnfeldman: so there is no hit

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsEntityIndexBatchImpl.java
shawnfeldman: i need to resolve all the promises though, not sure just resolving the observable will work without explicitly calling get

PR #163, opened 2015-02-20
updated to work with Laravel 5 and added the package to packagist.org 

apps4u:
This is a update so it will work with laravel 5 and I've also added the package to packagist.org so there is no need to add the repo url to your composer.json file any more its a full php composer package.


apps4u:
Sorry this is a update to the PHP 5 SDK only

GERey:
Hi I can't quite see what repo this PR came from so I can't accept the PR. If you could resubmit it from a valid repo I would be happy to accept the merge. ( Right now it just says unknown repository @apps4u )

PR #164, opened 2015-02-24
Usergrid 409

GERey:
Tested local deploy.
Added support for cassandra keyspaces to be set in the tomcat properties file.


stack/core/src/main/java/org/apache/usergrid/corepersistence/CpSetup.java
```diff
@@ -146,17 +145,17 @@ public void setupSystemKeyspace() throws Exception {
 
         migrate();
 
-        cass.createColumnFamily( SYSTEM_KEYSPACE,
-            createColumnFamilyDefinition( SYSTEM_KEYSPACE, APPLICATIONS_CF, ComparatorType.BYTESTYPE ) );
+        cass.createColumnFamily( getApplicationKeyspace(),
```
tnine: A lot of these CF's are necessary anymore.  Can you audit their usage and delete the ones that are no longer used?  IIRC the only ones we need to retain are the TOKENS_CF and the PRINCIPAL_TOKEN_CF for the inverse relationship

stack/core/src/main/java/org/apache/usergrid/corepersistence/CpSetup.java
```diff
@@ -202,17 +201,17 @@ public void setupStaticKeyspace() throws Exception {
 
         if ( USE_VIRTUAL_KEYSPACES ) {
```
tnine: This if isn't needed anymore, we don't always use virtual.  Every row key is app prefixed, and we put all systems into the same keyspace. Eventually, we'll move to an org->keyspace lookup system.

stack/core/src/main/java/org/apache/usergrid/persistence/cassandra/CassandraService.java
```diff
@@ -158,7 +164,7 @@ public void init() throws Exception {
         accessMap.put( "username", properties.getProperty( "cassandra.username" ) );
         accessMap.put( "password", properties.getProperty( "cassandra.password" ) );
         systemKeyspace =
-                HFactory.createKeyspace( SYSTEM_KEYSPACE, cluster, consistencyLevelPolicy, ON_FAIL_TRY_ALL_AVAILABLE,
+                HFactory.createKeyspace( getApplicationKeyspace() , cluster, consistencyLevelPolicy, ON_FAIL_TRY_ALL_AVAILABLE,
```
tnine: Can we remove hector calls to create keyspace and our system functions?  The CP modules are more intelligent in their creation of the keyspaces.  If we can remove it here, lets do that.  Note that hector can still create CFs, but the RF settings on the keyspaces seem to not work well in practice.

stack/core/src/main/java/org/apache/usergrid/persistence/cassandra/CassandraService.java
```diff
@@ -230,7 +239,7 @@ public void setConsistencyLevelPolicy( ConsistencyLevelPolicy consistencyLevelPo
     /** @return keyspace for application UUID */
     public static String keyspaceForApplication( UUID applicationId ) {
         if ( USE_VIRTUAL_KEYSPACES ) {
```
tnine: Just change this code, we always use virtual keyspaces, so the other scheme isn't valid any longer

stack/core/src/main/java/org/apache/usergrid/persistence/cassandra/SetupImpl.java
```diff
@@ -80,17 +79,17 @@ public void setupSystemKeyspace() throws Exception {
 
         logger.info( "Initialize system keyspace" );
```
tnine: Same comment as before

tnine:
We need to do some acceptance testing on this to ensure all this change does is invalidate tokens.

So we need to test the following.

# Deploy our previous release to a cluster
# Create an application.  Get our client_id and secret both org and app
# Deploy a build from this branch
# Ensure the client_id and secret for both org and app still work
# Ensure that we can generate a new token.  Test signups etc and everything works as expected.

GERey:
This change invalidates all previous tokens. 

Tested the following: 
Took previous release: Ran database/setup and superuser/setup  created a new user. Entered data into multiple collections. Saved the client_id and secret. Made sure I could get response from management/token using the client id and secret

Deployed new release: Had to rerun database/setup in order to recreate the token database.Made sure I could access all data from the user and multiple collections. Made hits against management/token and also <orgname>/<appname> with the previous client id and secret ( but with a new token ) and verified they still worked. Also took the tomcat offline and added in the following line to the properties file. 

cassandra.keyspace.application=test_keyspace

I then reran the database/setup and superuser/setup endpoints and created a user with the exact same email as the previous user. Then created data in the "test_keyspace". The tomcat was taken offline and switched back to the default keyspace and I logged in using the same email and found the data that was created in the previous release. The client id and secret are different for the keyspaces and won't work across keyspaces.

tnine:
Looks good. Merging.  Just a note, after this is deployed, there's no migration.  Users will need to re-log in to regenerate their tokens.  UG will now operate out of a singular keyspace.

PR #165, opened 2015-02-27
Usergrid 432: Add threads and cores back to pom.xml

shawnfeldman:
had to account for null issues on deleting apps.


tnine:
Looks good to me.

tnine:
Just tried to test this before merging and I'm unable to get all the test to pass. @shawnfeldman Can you see if they run on your box?

shawnfeldman:
added async deltebyquery

stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManagerFactory.java
```diff
@@ -317,12 +317,11 @@ public void deleteApplication(UUID applicationId) throws Exception {
...
         Results results = em.searchCollection( em.getApplicationRef(), "appinfos", q);
         Entity appToDelete = results.getEntity();
-        em.delete( appToDelete );
-
-        // create new Entity in deleted_appinfos collection, with same UUID and properties as deleted appinfo
-        em.create( "deleted_appinfo", appToDelete.getProperties() );
-
-        em.refreshIndex();
+        if(appToDelete != null) {
+            em.delete(appToDelete);
```
tnine: We should reverse this.  If the node dies after this call, we won't have marked it.  Better to have an erroneous mark than to miss it completely.

stack/core/src/main/java/org/apache/usergrid/corepersistence/CpWalker.java
```diff
@@ -58,8 +58,10 @@ public CpWalker(){
     public void walkCollections(final CpEntityManager em, final EntityRef start,
         String collectionName, boolean reverse, final CpVisitor visitor) throws Exception {
 
-        doWalkCollections(
-            em, collectionName, reverse, new SimpleId( start.getUuid(), start.getType() ), visitor );
+        if(start != null) {
```
tnine: Should we fail silently or blow up here?  Feels like we should do preconditions so we can flush out the incorrect caller impl

stack/rest/src/main/java/org/apache/usergrid/rest/IndexResource.java
```diff
@@ -250,8 +250,10 @@ public JSONWithPadding addIndex(@Context UriInfo ui,
             throw new IllegalArgumentException("Please add an indexSuffix to your post");
         }
 
+        String writeConsistency = config.get("writeConsistency") != null ? (String)config.get("writeConsistency") : "one" ;
```
tnine: Can we use IndexFig here?

PR #167, opened 2015-03-02
Fix for key names in Credentials.java

snoopdave:
Fixes many tests in ApplicationResourceIT.


tnine:
@snoopdave This looks good.  Can you just merge this into the Apache two-dot-o directly?

PR #169, opened 2015-03-03
Usergrid 280

GERey:
https://issues.apache.org/jira/browse/USERGRID-280

Also removed tomcat instance lying around in rest tier that is no longer user.


rodsimpson:
done!  Thanks!

PR #173, opened 2015-03-04
Fix for USERGRID-448: Remove redundant appinfos collections in ManagementServiceImpl

snoopdave:
DO NOT MERGE

Needs careful review.

Changes to merge appinfo and application_info collections into one application_info collection. Also:
- Removed unnecessary system and default applications
- Now use seek rather than search to load application info entities
- Uncommented the parts of ApplicationDeleteTest that did not work before the appinfo fix
- Will migrate old appinfo collection to application_info on startup, but can be configured to skip that

https://issues.apache.org/jira/browse/USERGRID-448


snoopdave:
Review notes
===
This is a big commit, so here are the key changes to consider.  Many files changed because I moved ApplicationInfo from the services module to the core module, my IDE did some uncalled for formatting (removing trailing spaces) and then I decided to move ApplicationInfo back home to services.

CpEntityManagerFactory
---
The main change here is that we no longer create an "appinfos" collection and instead rely solely on the "application_info" collection created by the ManagementServiceImpl.
   - System App and Default App have been removed completely, they were unused
   - No longer create redundant "organization" Entity
   - New impl. of deleteApplication() that saves connections
   - New impl. of restoreApplication() that restores connections
   - lookupAppication(name) uses alias instead of ES
   - migrationOldAppInfos() reads old "appinfo" entities and writes "application_info" entities. 
   - New property usergrid.twodoto.appinfo.migration which defaults to true

JobServiceScheduler
---
- Fixes to adapt to new way MetricsFactory is instantiated. Somebody changed it to be instantiated with Guice but did not fix this class.

ManagementServiceImpl
---
- New impl. of deleteApplication() that does some error checks and then calls emf.deleteApplication()
- New impl. of restoreApplication() calls emf.restoreApplication(), restores token and organization connection, and adds activity

ApplicationDeleteTest
---
- New and comprehensive app delete and restore tests

ApplicationsResource (test infrastructure)
---
- Represents management ApplicationsResource end-point

ManagementResponse (test infrastructure)
---
- Represents JSON response from management end-points

tnine:
Hey Dave.  Can you close this and point it at two-dot-o-dev?

snoopdave:
This has been merged it two-dot-o-dev

PR #178, opened 2015-03-09
Usergrid 465

GERey:


GERey:
https://issues.apache.org/jira/browse/USERGRID-465

PR #181, opened 2015-03-10
Usergrid 472 - this has been merged with 405

shawnfeldman:


stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsEntityIndexBatchImpl.java
```diff
@@ -255,7 +247,7 @@ private static Map entityToMap( EntityObject entity ) {
 
             if ( f instanceof ListField ) {
                 List list = ( List ) field.getValue();
-                entityMap.put( field.getName().toLowerCase(),
```
tnine: Should we use the same prefix for arrays and lists?  They have the same json representation, so it seems we'll want them to map the same in ES.

tnine:
I think we'll want to review this a bit more.  In my test merge, we're having issues with querying sub properties.   I'm not sure that when our target type in an array isn't a primitive, we want to prefix the array.  We'll need to identify the use cases, and mappings for this.

PR #182, opened 2015-03-11
Usergrid 473 - don't merge yet

shawnfeldman:


tnine:
Putting this on hold.  Our mapping is about to change, so no longer relevant.

PR #184, opened 2015-03-12
Usergrid 466

tnine:
Updated to the latest Codehale metrics

Removed unused annotations

Adds additional monitoring metrics.

Added a durable queue for indexing with using SQS as well as in memory queue.

Added a worker count for multi threaded queue consumption

Adds a dedicated master node to ES cluster.


tnine:
Don't ack yet, still working on polishing this branch

PR #187, opened 2015-03-18
Fixes issue with atomically moving indexes

tnine:


stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsEntityIndexImpl.java
```diff
@@ -261,60 +257,51 @@ public void addIndex(final String indexSuffix,final int numberOfShards, final in
...
-            isAck = adminClient.indices().prepareAliases().addAlias(
-                    indexName, alias.getReadAlias()).execute().actionGet().isAcknowledged();
-            timeAddReadAlias.stop();
-            logger.info("Created new read Alias Name [{}] ACK=[{}]", alias.getReadAlias(), isAck);
+            aliasesRequestBuilder.addAlias(  indexName, alias.getReadAlias());
+            logger.info("Created new read Alias Name [{}] on Index [{}]", alias.getReadAlias(), indexName);
+
 
-            //Added For Graphite Metrics
-            Timer.Context timeAddWriteAlias = addWriteAliasTimer.time();
```
GERey: Do these metrics not work? Otherwise why should we take them out?

tnine: They're gone.  We were doing 3 operations, now we do it in 1 round trip, so we only need 1 timer

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsEntityIndexImpl.java
```diff
@@ -261,60 +257,51 @@ public void addIndex(final String indexSuffix,final int numberOfShards, final in
...
-                        iane.getMessage() );
-                    continue;
-                }
-                finally {
-                    timeRemoveAlias.stop();
-                }
+            //remove the write alias from it's target
+            for ( String currentIndex : indexNames ) {
+                aliasesRequestBuilder.removeAlias( currentIndex, alias.getWriteAlias() );
+                logger.info("Removing existing write Alias Name [{}] from Index [{}]", alias.getReadAlias(), currentIndex);
```
GERey: If this removes the writeAlias, when shouldn't it be logging that it removed the alias.getWriteAlias instead of the alias.getReadAlias?

PR #191, opened 2015-03-19
Usergrid 493

GERey:
Fixed CollectionsResourceIT to logical state. One test still fails but that can be addressed later. 
Added generics to the namedResource so that we can start moving the tests over to generic posts.


GERey:
https://issues.apache.org/jira/browse/USERGRID-335 Issue that this resolves

GERey:
https://issues.apache.org/jira/browse/USERGRID-345 Also resolves the following issue as well.

GERey:
This PR is now included in #197

PR #193, opened 2015-03-20
USERGRID-457 - Add explicit Content-Length of 0 for POST request with no...

michaelarusso:
... body.


tnine:
Thanks!  I merged this into 2.0 and into 1.0

PR #199, opened 2015-03-24
Usergrid 501 - migration for single index

shawnfeldman:


stack/corepersistence/queryindex/src/test/java/org/apache/usergrid/persistence/index/impl/IndexMigrationTest.java
```diff
@@ -0,0 +1,106 @@
...
+
+            @Override
+            public void update(int migrationVersion, String message) {
+
+            }
+        };
+
+        TestIndexModule.TestAllApplicationsObservable obs = new TestIndexModule.TestAllApplicationsObservable(indexBatchBufferProducer,provider,indexCache,metricsFactory,fig);
+        int version = indexDataMigration.migrate(0, obs, po );
+        assertEquals(version, IndexDataVersions.SINGLE_INDEX.getVersion());
```
tnine: Can we read the indexes from ES and validate they have the read alias on them in this test too?

stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/migration/EsIndexDataMigrationImpl.java
```diff
@@ -0,0 +1,97 @@
...
+            return Observable.from(indexes);
+        })
+            .doOnNext(index -> {
+                IndicesAliasesRequestBuilder aliasesRequestBuilder = adminClient.indices().prepareAliases();
+                aliasesRequestBuilder = adminClient.indices().prepareAliases();
+                // add read alias
+                aliasesRequestBuilder.addAlias(index, indexIdentifier.getAlias().getReadAlias());
+                observer.update(latestVersion,"EsIndexDataMigrationImpl: fixed index: " + index );
+            })
+            .doOnError(error -> {
```
tnine: Lets only catch + swallow exceptions we expect.  Otherwise this will fail silently and the migration with succeed.

PR #202, opened 2015-03-27
USERGRID-457 - Update PingIdentityProvider to send parameters as formpar...

michaelarusso:
...ams instead of queryparams.


michaelarusso:
Already merged into the code base.  Closing.

PR #206, opened 2015-03-31
Usergrid 509

tnine:


tnine:
GH fail.  The UI didn't take the change.  Re-opening

PR #211, opened 2015-04-03
[USERGRID-538] - improve error response for uncaught errors

senthilkumar-kj:
Added UncaughtException to handle uncaught errors. It creates a time uuid and it is set as error id in error response. 

Sample response, 
> {"error":"uncaught","timestamp":1428023297834,"duration":1,"error_description":"Internal Server Error","exception":"org.apache.usergrid.rest.exceptions.UncaughtException","error_id":"e97cf251-d99d-11e4-8c2e-4eff6b02c98b"}


GERey:
LGTM thanks @senthilkumar-kj !

PR #213, opened 2015-04-07
Lookup applications using unique value instead of using query

snoopdave:
The fix is to lookup applications via unique value rather than query.

https://issues.apache.org/jira/browse/USERGRID-491
https://issues.apache.org/jira/browse/USERGRID-533


snoopdave:
The rest test is currently not working due to application_info UUID changes in two-dot-o-dev that impact REST layer authorization logic.

PR #216, opened 2015-04-08
Usergrid 550 - change migration to use new structure for app info

shawnfeldman:


stack/core/src/main/java/org/apache/usergrid/corepersistence/ApplicationIdCacheFactory.java
```diff
@@ -0,0 +1,43 @@
+/*
+ *
```
tnine: Since there's only a single instance of the cache, can we just use a Guice Provider?

shawnfeldman: its kind of a weird singleton since it needs and ecmf

tnine: Ahh, I see.  I missed that, it depends on the management app entity manager.

stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManagerFactory.java
```diff
@@ -347,67 +363,23 @@ public void deleteApplication(UUID applicationId) throws Exception {
                 }
             });
         }
-        final ApplicationEntityIndex aei = entityIndexFactory.createApplicationEntityIndex(deletedAppScope);
-        final Observable deleteNodeGraph = managementGraphManager.deleteNode(deletedAppId, Long.MAX_VALUE);
+        final ApplicationEntityIndex aei = entityIndexFactory.createApplicationEntityIndex(applicationScope);
+        final GraphManager managementGraphManager = managerCache.getGraphManager(managementAppScope);
+        final Observable deleteNodeGraph = managementGraphManager.deleteNode(applicationId, Long.MAX_VALUE);
```
tnine: Deleting the old app entity makes me nervous. In the event we encounter a bug, we wont' have any roll back strategy.  Can we just migrate to the new info, then possibly delete the old values in our next release after it's been deployed and we ensure it's solid?

shawnfeldman: yes

PR #225, opened 2015-04-15
Add hector settings

yding37:
Enable setting the following values for hector:
  - cassandraThriftSocketTimeout
  - useSocketKeepalive
Enabled HectorLockManager

### DIFF
--- a/stack/config/src/main/resources/usergrid-default.properties
+++ b/stack/config/src/main/resources/usergrid-default.properties
@@ -62,6 +62,9 @@ cassandra.mutation.flushsize=2000
 #Note that if this is deployed in a production cluster, the RF on the keyspace MUST be updated to use an odd number for it's replication Factor.
 #Even numbers for RF can potentially case the locks to fail, via "split brain" when read at QUORUM on lock verification
 cassandra.lock.keyspace=Locks
+#locking read & write policies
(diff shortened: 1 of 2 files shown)


stack/config/src/main/resources/usergrid-default.properties
```diff
@@ -62,6 +62,15 @@ cassandra.mutation.flushsize=2000
 #Note that if this is deployed in a production cluster, the RF on the keyspace MUST be updated to use an odd number for it's replication Factor.
 #Even numbers for RF can potentially case the locks to fail, via "split brain" when read at QUORUM on lock verification
 cassandra.lock.keyspace=Locks
+#locking read & write policies
+cassandra.lock.readcl=QUORUM
```
snoopdave: both of these should be local quorum by default

snoopdave:
what version of Hector adds the new cassandraThriftSocketTimeout parameters?

yding37:
I'm not too sure actually, although the version that's shipped with
usergrid has that setting.

On Thu, Apr 16, 2015 at 12:42 PM, David M. Johnson <notifications@github.com

yding37:
is this necessary for build purposes?

GERey:
Hello! If you could update this PR to address the conflicts I would be more than happy to accept the PR. Thanks!

PR #227, opened 2015-04-17
Per standard policy, we can't over write the repos.  Was causing the bui...

johnament:
...lds to not deploy.

I opened an infra ticket, they pointed out the issue.  Do you guys use this to deploy internally?  If so we'll need to review as it conflicts with ASF policy.


johnament:
Anyone want to take a look?

tnine:
@snoopdave This one is all you.  I haven't touched this in a while.

PR #228, opened 2015-04-17
Usergrid 409 1.0

GERey:


snoopdave:
Also need to be able to configure the Locks keyspace used in HectorLockManagerImpl.

Currently it is hardcoded into our Spring config.

GERey:
I was under the impression that this was already configurable by using "cassandra.lock.keyspace". Also in the usergrid-core-context.xml the HectorLockManagerImpl is already commented out so changing this value in the spring config wouldn't do anything.

PR #229, opened 2015-04-17
Added elastic search timeout to search calls.

GERey:
DO NOT MERGE
FOR REVIEW


tnine:
Looks good to me.  Can we merge into two-dot-o-dev as well?

PR #230, opened 2015-04-17
Ensure that when external token validation is enabled, Admin Users cannot sign-up

snoopdave:
Or be activated or be confirmed... all that should happen via Usergrid Central.

### DIFF
--- a/stack/rest/src/main/java/org/apache/usergrid/rest/management/users/UserResource.java
+++ b/stack/rest/src/main/java/org/apache/usergrid/rest/management/users/UserResource.java
@@ -393,7 +393,7 @@ public class UserResource extends AbstractContextResource {
                 !StringUtils.isEmpty( properties.getProperty( ManagementResource.USERGRID_CENTRAL_URL ) );
 
         if ( externalTokensEnabled ) {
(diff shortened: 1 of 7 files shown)


stack/rest/src/main/java/org/apache/usergrid/rest/management/users/UserResource.java
```diff
@@ -355,6 +389,14 @@ public JSONWithPadding reactivate( @Context UriInfo ui,
                                        @QueryParam( "callback" ) @DefaultValue( "callback" ) String callback )
             throws Exception {
 
+        final boolean externalTokensEnabled =
+                !StringUtils.isEmpty( properties.getProperty( ManagementResource.USERGRID_CENTRAL_URL ) );
+
+        if ( externalTokensEnabled ) {
+            throw new IllegalArgumentException( "Admin Users must reactiveate via " +
```
GERey: typo

GERey:
lgtm

GERey:
Just for future reference, issue is detailed here https://issues.apache.org/jira/browse/USERGRID-575

PR #235, opened 2015-04-21
fix rest compilation failures

senthilkumar-kj:
Fixed compilation failures in rest module. Added index methods to EntityManagerFactory to impl later.

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/persistence/EntityManagerFactory.java
+++ b/stack/core/src/main/java/org/apache/usergrid/persistence/EntityManagerFactory.java
@@ -17,15 +17,14 @@
 package org.apache.usergrid.persistence;
 
 
-import java.util.Map;
-import java.util.UUID;
-
 import com.google.common.base.Optional;
-
 import org.apache.usergrid.corepersistence.index.ReIndexService;
 import org.apache.usergrid.persistence.core.util.Health;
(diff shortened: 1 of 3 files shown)


stack/core/src/main/java/org/apache/usergrid/persistence/EntityManagerFactory.java
```diff
@@ -189,6 +188,12 @@ public UUID initializeApplication(
 
     public Health getIndexHealth();
 
+    void rebuildAllIndexes(ProgressObserver progressObserver);
```
tnine: I could be wrong, but I thought I already removed these today in the 2.0-dev branch. Can you pull and re-merge?  I'm going to reference the indexService directly in our rest tier, rather than continue to funnel data through the emf interfaces.  Eventually we'll have more events in the refactored system.

senthilkumar-kj: hi @tnine, I still see the line here - https://github.com/apache/incubator-usergrid/blob/two-dot-o-dev/stack/rest/src/main/java/org/apache/usergrid/rest/IndexResource.java#L88
I'll pull and push again

GERey:
Hi! These changes were addressed and compilation issues are fixed. If you don't mind could you close or update the pr.

PR #236, opened 2015-04-21
[USERGRID-527] - external load test for ES

senthilkumar-kj:
Please share your comments, so that I can update the tests to suit the requirement.


senthilkumar-kj:
I'm using the json available in the tests. I would like to change it to the one from s3 like woolworths. Will be working on it shortly.

senthilkumar-kj:
Build generates a zip file queryindex-2.0.0-SNAPSHOT-elasticsearch-distribution-test.zip. The script can be run on multiple instances (with same config) to hit ES similar to how UG does.

GERey:
Hi. We've developed and created some python scripts that have equal functionality. If you don't mind could you update or close this PR? Thanks!

PR #246, opened 2015-05-13
Usergrid 607

GERey:
https://issues.apache.org/jira/browse/USERGRID-607

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/corepersistence/index/IndexServiceImpl.java
+++ b/stack/core/src/main/java/org/apache/usergrid/corepersistence/index/IndexServiceImpl.java
@@ -20,6 +20,8 @@
 package org.apache.usergrid.corepersistence.index;
 
 
+import java.util.Iterator;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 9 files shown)


stack/core/src/main/java/org/apache/usergrid/corepersistence/index/IndexServiceImpl.java
```diff
@@ -163,16 +168,54 @@ public IndexServiceImpl( final GraphManagerFactory graphManagerFactory, final En
...
-        //TODO, query ES and remove this edge
 
-        throw new NotImplementedException( "Implement me" );
-    }
+                //review why generating the Scope from the Source  and the target node makes sense.
+                final IndexEdge fromSource = generateScopeFromSource( edge );
+                final Id targetId = edge.getTargetNode();
+
+
+                CandidateResults targetEdgesToBeDeindexed = ei.getAllEdgeDocuments( fromSource, targetId, 1000, 0 );
```
tnine: Can you instead refactor this to use Functional?  This way we have 1 type of implementation, and the codebase becomes more consistent.

GERey:
This is being rolled into usergrid-608 since there are changes there that affect the scope of USERGRID-607

PR #248, opened 2015-05-15
Usergrid 643

tnine:
Implements new IndexRebuild services


tnine:
This is ready to merge.  There are still issues in the /system/database/setup bootstrap.  This will need further testing before release.

stack/core/src/main/java/org/apache/usergrid/corepersistence/asyncevents/SQSAsyncEventService.java
```diff
@@ -184,7 +185,7 @@ public void ack( final List<QueueMessage> messages ) {
     }
 
 
-    @Override
+//    @Override
     public void index( final EntityIdScope entityIdScope ) {
```
GERey: Seems like we're tearing this method out of the InMemoryAsyncEvenService, why keep it here ?

stack/core/src/main/java/org/apache/usergrid/corepersistence/index/IndexProcessorFig.java
```diff
@@ -78,9 +78,9 @@
     String getQueueImplementation();
 
 
-    @Default("30000")
-    @Key("elasticsearch.reindex.sample.interval")
-    long getReIndexSampleInterval();
+    @Default("1000")
```
GERey: Is there a reason we think a second is optimal?

stack/core/src/main/java/org/apache/usergrid/corepersistence/index/ReIndexServiceImpl.java
```diff
@@ -70,60 +79,228 @@
...
+
+
+        final CursorSeek<Edge> cursorSeek = getResumeEdge( cursor );
+
+        final Optional<ApplicationScope> appId = reIndexRequestBuilder.getApplicationScope();
 
-        final Observable<ApplicationScope>  applicationScopes = appId.isPresent()? Observable.just( getApplicationScope(appId.get()) ) : allApplicationsObservable.getData();
 
-        final String newCursor = StringUtils.sanitizeUUID( UUIDGenerator.newTimeUUID() );
+        Preconditions.checkArgument( !(cursor.isPresent() && appId.isPresent()),
```
GERey: Why not catch this in the builder?

stack/core/src/main/java/org/apache/usergrid/corepersistence/index/ReIndexServiceImpl.java
```diff
@@ -70,60 +79,228 @@
...
+            .subscribeOn( Schedulers.io() ).subscribe();
+
+
+        return new ReIndexStatus( jobId, Status.STARTED, 0, 0 );
+    }
+
+
+    @Override
+    public ReIndexRequestBuilder getBuilder() {
+        return new ReIndexRequestBuilderImpl();
```
GERey: Should this point to the builder rather than the impl itself , then wire the impl up through guice?

stack/core/src/test/java/org/apache/usergrid/persistence/PerformanceEntityRebuildIndexTest.java
```diff
@@ -22,207 +22,149 @@
...
     public void printReport() {
-        logger.debug("Printing metrics report");
-        reporter.report();
-        reporter.stop();
+        logger.debug( "Printing metrics report" );
     }
 
 
-    @Test
+    @Test( timeout = 120000 )
```
GERey: Seems like a timeout this long isn't really all that helpful. By this time somebody would have probably cut short the test.

stack/core/src/test/java/org/apache/usergrid/persistence/PerformanceEntityRebuildIndexTest.java
```diff
@@ -231,82 +173,79 @@ public void onProgress( final EntityRef entity ) {
...
```
GERey: same as above

stack/core/src/test/java/org/apache/usergrid/persistence/PerformanceEntityRebuildIndexTest.java
```diff
@@ -317,61 +256,73 @@ public void rebuildIndex() throws Exception {
...
+        readData( em, collectionName, ENTITIES_TO_INDEX, 3 );
+    }
+
+
+    /**
+     * Wait for the rebuild to occur
+     */
+    private void waitForRebuild( final ReIndexService.ReIndexStatus status, final ReIndexService reIndexService )
+        throws InterruptedException {
+        while ( true ) {
```
GERey: Isn't it possible that this could take forever? Maybe the timeout should be placed here?

stack/core/src/test/java/org/apache/usergrid/persistence/PerformanceEntityRebuildIndexTest.java
```diff
@@ -317,61 +256,73 @@ public void rebuildIndex() throws Exception {
...
+
+            try {
+                final ReIndexService.ReIndexStatus updatedStatus = reIndexService.getStatus( status.getJobId() );
+
+                if ( updatedStatus.getStatus() == ReIndexService.Status.COMPLETE ) {
+                    break;
+                }
+            }
+            catch ( IllegalArgumentException iae ) {
+                //swallow
```
GERey: Would be nice for a comment to say why its ok to have IllegalArgumentExceptions thrown somewhere.

stack/core/src/main/java/org/apache/usergrid/corepersistence/index/ReIndexServiceImpl.java
tnine: A good point.  We can catch it in either place.  I personally just prefer to keep validation close to the logic, no other reason.

stack/core/src/main/java/org/apache/usergrid/corepersistence/index/ReIndexServiceImpl.java
tnine: Nah, Builder is just a bean, so we'll hide it's generation in the service, this way if it ultimately needs dependencies, we can wire them as required.

stack/core/src/test/java/org/apache/usergrid/persistence/PerformanceEntityRebuildIndexTest.java
tnine: Yeah, agreed.  This is because I encountered a bug where we were infinitely indexing.  This is just a failsafe in case that happens again.

stack/core/src/test/java/org/apache/usergrid/persistence/PerformanceEntityRebuildIndexTest.java
tnine: Yes, it could, but the test will timeout first due to the timeout annotation.

PR #250, opened 2015-05-18
Upgrade JClouds, add connection pool for Central SSO

snoopdave:
This PR upgrades from to JClouds 1.7.1 and because JClouds no longer supports async calls to AWS, I had to change code in S3BinaryStore to create a thread for large uploads (i.e. larger than 5mb).

This PR also includes changes from the branch sso-conn-pool, which adds a connection pool for Central SSO HTTPClient calls. See the PR #245 if you want to review those changes separately.

### DIFF
--- a/stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
+++ b/stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
@@ -25,10 +25,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.util.Map;
+import java.util.Properties;
+import java.util.Stack;
 import java.util.UUID;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Executors;
+import java.util.concurrent.*;
 
+import org.apache.usergrid.utils.StringUtils;
 import org.jclouds.ContextBuilder;
 import org.jclouds.blobstore.AsyncBlobStore;
 import org.jclouds.blobstore.BlobStore;
@@ -63,12 +65,16 @@ public class S3BinaryStore implements BinaryStore {
 
     private static final Logger LOG = LoggerFactory.getLogger( S3BinaryStore.class );
     private static final long FIVE_MB = ( FileUtils.ONE_MB * 5 );
+    private static String WORKERS_PROP_NAME = "usergrid.binary.upload-workers";
 
     private BlobStoreContext context;
     private String accessId;
     private String secretKey;
     private String bucketName;
-    private ExecutorService executor = Executors.newFixedThreadPool( 10 );
+    private ExecutorService executorService;
+
+    @Autowired
+    private Properties properties;
 
     @Autowired
     private EntityManagerFactory emf;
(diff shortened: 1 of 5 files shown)


stack/rest/src/main/java/org/apache/usergrid/rest/management/ManagementResource.java
```diff
@@ -701,6 +710,50 @@ private JsonNode getMeFromUgCentral( String extAccessToken )  throws EntityNotFo
     }
 
 
+    synchronized Client getJerseyClient() {
+
+        if ( jerseyClient == null ) {
```
tnine: Should we put synchronized in the null block? This way everyone won't have to queue up to get the client once it's instantiated.

stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
```diff
@@ -134,10 +134,11 @@ public void write( final UUID appId, final Entity entity, InputStream inputStrea
         }
         else { // bigger than 5mb... dump 5 mb tmp files and upload from them
 
-            // todo: yes, AsyncBlobStore is deprecated, but there appears to be no replacement yet
-            final AsyncBlobStore blobStore = getContext().getAsyncBlobStore();
+            // create temp file and copy entire file to that temp file
 
-            File tempFile = File.createTempFile( entity.getUuid().toString(), "tmp" );
+            LOG.debug( "Writing temp file for S3 upload" );
```
tnine: This code itself looks great.  However I noticed on the total impl we don't have a max file size we support.  Since we're going to have to upload this to the tomcat server, should we enforce a max file size to stop someone from overloading the tomcats?

stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
```diff
@@ -149,38 +150,52 @@ public void write( final UUID appId, final Entity entity, InputStream inputStrea
...
 
-            Runnable listener = new Runnable() {
+            LOG.debug( "Starting upload thread" );
+
+            Thread uploadThread = new Thread( new Runnable() {
                 @Override
                 public void run() {
                     try {
-                        String eTag = future.get();
+                        LOG.debug( "S3 upload thread started" );
```
tnine: Can we use an executor pool with a queue and a callable?  This way we wont' create a lot of threads that may use our file handles during uploads.

stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
snoopdave: Todd, it appears that we (unintentionally?) limit file size to 10 mb, because we read 5pm and then on this line we only only read up to 5 pm more: https://goo.gl/sf0Tof

rgex:
I agree with tnine that we should add a max file size parameter that is configurable in the config file.
If configuring a max file size using tomcat we wont get a nice JSON error message.

stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
rgex: I agree with tnine that we should add a max file size parameter that is configurable in the config file.
If configuring a max file size using tomcat we wont get a nice JSON error message.

stack/rest/src/main/java/org/apache/usergrid/rest/management/ManagementResource.java
snoopdave: done

stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
snoopdave: done

stack/services/src/main/java/org/apache/usergrid/services/assets/data/S3BinaryStore.java
snoopdave: I added a new property 'usergrid.binary.max-size-mb' that limits the upload size and it defaults to 50mb.

PR #259, opened 2015-06-02
Fix logic issue when recaptcha is disabled

r3b:


GERey:
Merged! If you could delete this branch it would be most appreciated!

PR #263, opened 2015-06-04
new name strategy

shawnfeldman:
Update Index prefix for management app to use the cassandra keyspace name


stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/IndexIdentifierv2Impl.java
```diff
@@ -0,0 +1,85 @@
...
+    private final IndexFig config;
+    private final String prefix;
+
+    @Inject
+    public IndexIdentifierv2Impl(final Keyspace keyspace, final IndexFig config) {
+        this.config = config;
+        this.prefix = getPrefix(keyspace, config);
+    }
+
+    private String getPrefix(Keyspace keyspace, IndexFig config) {
```
tnine: Can we just take the Fig or a string here instead of the keyspace itself?  This way our index won't depend on Cass libraries, just a name.

PR #264, opened 2015-06-04
[USERGRID-613] Fixed errors with running index endpoint.

GERey:
Added testing compatibility in rest test framework for index tests. (i.e adding a call for basic auth )
Added debugging for finding out if you are a superuser or not.
Added test for rebuilding endpoint and ignored previous unused test as fixing and adding tests for that should be a separate issue.

### DIFF
--- a/stack/rest/src/main/java/org/apache/usergrid/rest/IndexResource.java
+++ b/stack/rest/src/main/java/org/apache/usergrid/rest/IndexResource.java
@@ -41,6 +41,7 @@ import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Component;
 
 import org.apache.usergrid.corepersistence.index.ReIndexRequestBuilder;
+import org.apache.usergrid.corepersistence.index.ReIndexRequestBuilderImpl;
 import org.apache.usergrid.corepersistence.index.ReIndexService;
 import org.apache.usergrid.persistence.index.utils.UUIDUtils;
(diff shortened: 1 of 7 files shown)


stack/rest/src/main/java/org/apache/usergrid/rest/IndexResource.java
```diff
@@ -250,7 +252,8 @@ private ReIndexService getReIndexService() {
 
 
     private ReIndexRequestBuilder createRequest() {
-        return createRequest();
+        //TODO: wire this up through spring, and in the future guice.
+        return new ReIndexRequestBuilderImpl();
```
GERey: This was an endless loop, and should be returning the builder this is making a call through to core skipping services. Maybe this needs a ticket associated with it?

stack/rest/src/test/java/org/apache/usergrid/rest/test/resource2point0/endpoints/NamedResource.java
```diff
@@ -248,6 +250,29 @@ public ApiResponse  post(boolean useToken, Map map, QueryParameters queryParamet
 
     }
 
+    //Used for empty posts
+    public <T> T post( boolean useToken, Class<T> type, Map entity, final QueryParameters queryParameters, boolean useBasicAuthentication ) {
```
GERey: Overloaded the existing method to save effort on refactoring it all. Should go away when we tear the tokens out of the post calls as authorization will be handled seperately.

PR #265, opened 2015-06-04
[USERGRID-613] Added fix mixed cases in two-dot-o.

GERey:


stack/rest/src/main/java/org/apache/usergrid/rest/IndexResource.java
```diff
@@ -178,7 +178,7 @@ public JSONWithPadding rebuildIndexes(
 
             public void run() {
 
-                logger.info( "Started rebuilding application {} in collection {}", appId, collectionName );
```
shawnfeldman: why not just tolowercase once?

PR #266, opened 2015-06-05
[USERGRID-671] Adding test that proves that this issue has already be…

GERey:
…en resolved.

https://issues.apache.org/jira/browse/USERGRID-671


shawnfeldman:
looks good

PR #267, opened 2015-06-05
renamed JMX counters/timers, moved some to be instance attributes vs …

jwest-apigee:
…local attributes, defined in method


jwest-apigee:
don't use...

PR #268, opened 2015-06-05
Small fixes to Admin import/export tool based on results of testing

snoopdave:
- ExportAdmins now does not halt when getOrganizationsForAdminUser() has uniqueness violation
- ImportAdmins logic now only creates an organization if it does not already exist
- Additional test assertion to cover managementService.getAdminUsersForOrganization()


GERey:
Looks good to me. Thanks Dave!

PR #269, opened 2015-06-05
Updated timer names and added .timer and .meter to the factory to create them

jwest-apigee:


stack/services/src/main/java/org/apache/usergrid/services/notifications/impl/ApplicationQueueManagerImpl.java
```diff
@@ -175,13 +179,13 @@ public void queueNotification(final Notification notification, final JobExecutio
...
-                                        .flatMap( entity -> Observable.just( entity ).map( entityListFunct )
-                                                                      .doOnError( throwable -> {
-                                                                          LOG.error( "Failed while writing",
-                                                                              throwable );
-                                                                      } ).subscribeOn( Schedulers.io() )
+                                        .flatMap(entity -> Observable.just(entity).map(entityListFunct)
+                                            .doOnError(throwable -> {
+                                                LOG.error("Failed while writing",
+                                                    throwable);
+                                            }).subscribeOn(Schedulers.io())
```
tnine: Since we block on line 190, we won't need to subscribe on Schedulers.io().  I know you didn't add this, but can we remove it since we're spinning up another thread that we don't need?

tnine:
LGTM.  I'll ack this later today when I merge other PRs

PR #270, opened 2015-06-05
Added ability for application client resources to be accepted.

GERey:
Also added additional test. 
Additional details below:
https://issues.apache.org/jira/browse/USERGRID-670

### DIFF
--- a/stack/services/src/main/java/org/apache/usergrid/security/shiro/utils/SubjectUtils.java
+++ b/stack/services/src/main/java/org/apache/usergrid/security/shiro/utils/SubjectUtils.java
@@ -250,12 +250,7 @@ public class SubjectUtils {
         }
         if ( identifier.isName() ) {
             applicationName = identifier.getName().toLowerCase();
-            for (String value: values ){
-                if(value.toLowerCase().equals( applicationName )){
-                    applicationId = applications.inverse().get( value );
-                    break;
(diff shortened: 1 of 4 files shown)


stack/services/src/main/java/org/apache/usergrid/security/shiro/utils/SubjectUtils.java
```diff
@@ -243,14 +243,18 @@ public static ApplicationInfo getApplication( Identifier identifier ) {
...
+
         if ( applications == null ) {
             return null;
         }
         if ( identifier.isName() ) {
             applicationName = identifier.getName().toLowerCase();
-            applicationId = applications.inverse().get( applicationName );
-            if ( applicationId == null ) {
-                applicationId = applications.inverse().get( identifier.getName() );
+            for (String value: values ){
```
tnine: Dumb question, but why not just get from the bi-map after normalizing the string.  Then we'll have an O(1) lookup vs O(n)?

GERey: The reason I normalize the string is just defensive, the real reason I need to normalize the string is because the value that gets created in the Bimap from getApplications is mixed case and it get instantiated through a Session object. Thats why I take this approach.

PR #277, opened 2015-06-11
Improve Export Admins performance.

snoopdave:
Use multiple threads to read data from Cassandra and one thread to write that data to the two output files. How it works:

- In main thread we query for IDs of all admin users, add each ID to read queue.
- Read-queue workers read admin user data, add data to write queue.
- One write-queue worker reads data writes to file.

Depending on number of threads you specify with the new -readThreads parameter performance is increased by as much as 10X.

### DIFF
--- a/stack/tools/src/main/java/org/apache/usergrid/tools/ExportAdmins.java
+++ b/stack/tools/src/main/java/org/apache/usergrid/tools/ExportAdmins.java
@@ -337,7 +337,6 @@ public class ExportAdmins extends ExportingToolBase {
 
                     // write user to application file
                     usersFile.writeObject( task.adminUser );
-                    //usersFile.writeEndObject();
                     echo( task.adminUser );
 
                     // write metadata to metadata file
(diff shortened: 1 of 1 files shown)


stack/tools/src/main/java/org/apache/usergrid/tools/ExportAdmins.java
```diff
@@ -60,213 +82,393 @@ public void runTool( CommandLine line ) throws Exception {
...
+            // write one JSON file for management application users
+            JsonGenerator usersFile =
+                    getJsonGenerator( createOutputFile( ADMIN_USERS_PREFIX, em.getApplication().getName() ) );
+            usersFile.writeStartArray();
 
-        Set<String> dictionaries = em.getDictionaries( entity );
-        for ( String dictionary : dictionaries ) {
+            // write one JSON file for metadata: collections, connections and dictionaries of those users
+            JsonGenerator metadataFile =
+                    getJsonGenerator( createOutputFile( ADMIN_USER_METADATA_PREFIX, em.getApplication().getName() ) );
```
GERey: So we're using the application's name to name the output file for the admin user metadata? Wouldn't this lead to conflicts in the case that multiple admin users could belong to the same application?

snoopdave: No, because all Admin Users belong to the one and only management app.

PR #282, opened 2015-06-15
Usergrid 751 - change from smile to json serializer, use string serialization 

shawnfeldman:


stack/corepersistence/collection/src/main/java/org/apache/usergrid/persistence/collection/serialization/impl/MvccEntitySerializationStrategyV3Impl.java
```diff
@@ -458,9 +460,9 @@ public ByteBuffer toByteBuffer( final EntityWrapper wrapper ) {
             final Entity entity = wrapper.entity.get();
             //Convert to internal entity map
             final EntityMap entityMap = EntityMap.fromEntity( entity );
-            final byte[] entityBytes;
```
tnine: This still seems to use the fieldbuffer byte buffer.  Can we remove this completely and just use a nested JSON object?

shawnfeldman: i think it was removed

stack/corepersistence/collection/src/main/java/org/apache/usergrid/persistence/collection/serialization/impl/MvccEntitySerializationStrategyV3Impl.java
```diff
@@ -458,9 +460,9 @@ public ByteBuffer toByteBuffer( final EntityWrapper wrapper ) {
...
+            final String entityString;
             try {
-                entityBytes = MAPPER.writeValueAsBytes( entityMap );
+                entityString = MAPPER.writeValueAsString( entityMap );
```
tnine: This still seems to use the fieldbuffer byte buffer.  Can we remove this completely and just use a nested JSON object?

PR #284, opened 2015-06-16
remove default map properties

shawnfeldman:


tnine:
LGTM.  As long as all the tests pass, we should merge.

PR #287, opened 2015-06-18
Fix test problems and re-enable AssetResourceIT tests

snoopdave:


stack/rest/src/test/java/org/apache/usergrid/rest/applications/assets/AssetResourceIT.java
```diff
@@ -111,8 +109,10 @@ public void octetStreamOnDynamicEntity() throws Exception {
 
         String orgAppPath = clientSetup.getOrganizationName() + "/" + clientSetup.getAppName();
 
-        JsonNode node = mapper.readTree( resource().path( orgAppPath + "/foos" ).queryParam( "access_token", access_token )
-            .accept( MediaType.APPLICATION_JSON ).type( MediaType.APPLICATION_JSON_TYPE )
+        JsonNode node = mapper.readTree( resource().path( orgAppPath + "/foos" )
```
shawnfeldman: should convert this to new structure if possible, conversion is pretty straightforward

PR #288, opened 2015-06-19
Updates to SNS/SQS but deserialization is not working right

jwest-apigee:


jwest-apigee:
Serialization update with Russo's SNS envelope handling.  Messages are not getting ack'd to the queue.  RX Setup may not be right?

jwest-apigee:
closing and changing title

PR #295, opened 2015-06-30
Adding test proving the issue and added a fix to the mapping for elas…

GERey:
…tic search so the issue should be fixed.


tnine:
LGTM. Feel free to merge

PR #296, opened 2015-07-01
isolating app info from old app infos, so we can do in place migration

shawnfeldman:


shawnfeldman:
lets change app infos, fixed migration so it creates things the "right" way.  changed app infos

tnine:
LGTM.

PR #300, opened 2015-07-07
Ensure that the input application name does not contain characters th…

arun-ram:
…at can be used to inject malicious content.

The application name is repeated as-is in the response message if the oragnization/application cannot be found when OrganizationApplicationNotFoundException is thrown. This could be used for an XSS attack.

### DIFF
--- a/stack/rest/src/main/java/org/apache/usergrid/rest/organizations/OrganizationResource.java
+++ b/stack/rest/src/main/java/org/apache/usergrid/rest/organizations/OrganizationResource.java
@@ -125,14 +125,15 @@ public class OrganizationResource extends AbstractContextResource {
     }
 
 
-    @Path("{applicationName: [^<>/]+}")
+    @Path("{applicationName}")
     public ApplicationResource getApplicationByName( @PathParam("applicationName") String applicationName )
             throws Exception {
-
         if ( "options".equalsIgnoreCase( request.getMethod() ) ) {
             throw new NoOpException();
         }
-
+        if (!isSafe(applicationName)) {
+            throw new IllegalArgumentException("Invalid application name");
+        }
         String orgAppName = PathingUtils.assembleAppName( organizationName, applicationName );
         UUID applicationId = emf.lookupApplication( orgAppName );
         if ( applicationId == null ) {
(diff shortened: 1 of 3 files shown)


rgex:
Why not use a whitelist such as [-_A-Za-z0-9] ?

arun-ram:
I used an exclusion list rather than a white list to allow for non-English characters in the application name. If other languages are not supported, we could change it to a white list.

stack/rest/src/main/java/org/apache/usergrid/rest/organizations/OrganizationResource.java
```diff
@@ -143,21 +143,21 @@ public ApplicationResource getApplicationByName( @PathParam("applicationName") S
     }
 
 
-    @Path("applications/{applicationName}")
+    @Path("applications/{applicationName: [^<>/]+}")
```
johnament: I tested this out locally, not sure if it's doing what you'd expect.  I went to this URL (my local image shows 8580 as the HTTP port)

http://localhost:8580/applications/%3Cscript%3Ealert('bob')%3C/script%3E

The response still included the script tag, so this came back

```

{"error":"organization_application_not_found","timestamp":1436323964157,"duration":3,"exception":"org.apache.usergrid.rest.exceptions.OrganizationApplicationNotFoundException","error_description":"Could not find application for applications/<script>alert('bob')< from URI: applications/<script>alert('bob')</script>"}
```

rgex:
Collections are also affected

stack/rest/src/main/java/org/apache/usergrid/rest/organizations/OrganizationResource.java
johnament: If you want to understand why the output always includes the unescaped URI, take a look at [OrganizationApplicationNotFoundException.java](https://github.com/apache/incubator-usergrid/blob/master/stack/rest/src/main/java/org/apache/usergrid/rest/exceptions/OrganizationApplicationNotFoundException.java#L38)

rgex:
From RFC 3986:
   2.3.  Unreserved Characters
   Characters that are allowed in a URI but do not have a reserved
   purpose are called unreserved.  These include uppercase and lowercase
   letters, decimal digits, hyphen, period, underscore, and tilde.

      unreserved  = ALPHA / DIGIT / "-" / "." / "_" / "~"

So an organisation/application/collection should only contain this characters.

arun-ram:
Based on feedback from rgex, I am now using a whitelist that includes the unreserved characters. I found that I cannot use JAX-RS path param regexp URI matching because the matching is done before URI decoding.

stack/rest/src/test/java/org/apache/usergrid/rest/management/organizations/OrganizationResourceIT.java
```diff
@@ -32,7 +35,9 @@
 import org.apache.usergrid.rest.TestContextSetup;
 
 import junit.framework.Assert;
+import org.usergrid.java.client.response.ApiResponse;
 
+import static junit.framework.Assert.fail;
```
johnament: Could you use org.junit here?

arun-ram: done

rgex:
I would suggest that we check the organization, application, collection and connection names.
Similar to CouchDB I think that we should allow ALPHA / DIGIT / "-" / "." / "_" / "~" with one exception: names starting with underscore would be system reserved.

arun-ram:
I see that Lynch Lee is working on USERGRID-868 and has a pull request open for validating entity names. @rgex Should we merge this pull request and let Lynch Lee take care of collection and connection names?

johnament:
I've merged this, if there's more content to add please raise more PRs.  The more the merrier.

Thanks for the contrib @arun-ram !

PR #302, opened 2015-07-08
Usergrid 752

tnine:
Updates the search engine to use concurrent shard searching


tnine:
Do not merge yet.  I need to perform more integration testing with existing clusters.

snoopdave:
This is a huge change. It might be a good idea to do a screen-sharing session with a couple of UG devs and walk through the design and code.

tnine:
Yeah, I agree, a peer review would be ideal.  We're still doing performance testing.  Once we're happy with that, we should review everything.  I want to make sure this is performant before we push it through.

PR #307, opened 2015-07-15
New ExportApp tool, using RxJava

snoopdave:
https://issues.apache.org/jira/browse/USERGRID-788

### DIFF
--- a/stack/tools/src/main/java/org/apache/usergrid/tools/ExportApp.java
+++ b/stack/tools/src/main/java/org/apache/usergrid/tools/ExportApp.java
@@ -1,0 +1,687 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


stack/tools/src/main/java/org/apache/usergrid/tools/ExportApp.java
```diff
@@ -0,0 +1,552 @@
...
+        Observable<String> collectionsObservable = Observable.create( new CollectionsObservable( em ) );
+        
+        collectionsObservable.flatMap( new Func1<String, Observable<ExportEntity>>() {
+            
+            public Observable<ExportEntity> call(String collection) {
+                return Observable.create( new EntityObservable( em, collection ))
+                        .doOnNext( new EntityWriteAction() ).subscribeOn( writeScheduler );
+            }
+            
+        }, 10).flatMap( new Func1<ExportEntity, Observable<ExportConnection>>() {
```
tnine: Rather than have a static 10 here, we should probably use the Schedulers.io() scheduler, then use the write thread count as this value. Otherwise you'll just introduce contention on a thread pool.  If we use the unbounded Schedulers.io(), and us the size as the flatmap argument, we'll never exceed the set thread size.

stack/tools/src/main/java/org/apache/usergrid/tools/ExportApp.java
```diff
@@ -0,0 +1,552 @@
...
+            
+            public Observable<ExportConnection> call(ExportEntity exportEntity) {
+                return Observable.create( new ConnectionsObservable( em, exportEntity ))
+                        .doOnNext( new ConnectionWriteAction() ).subscribeOn( writeScheduler );
+            }
+            
+        }, 10)
```
tnine: Same as above

snoopdave:
@tnine Please take another look. I made flatmap max observables match write thread count and use Schedulers.io() instead of a custom readScheduler.  But I had to keep the writeScheduler in place, otherwise my test fails because (I think) threads explode and Cassandra falls over.

PR #308, opened 2015-07-15
Import entities parallel

tnine:
Allows us to import entities in parallel, drastically increasing response times


michaelarusso:
:+1:

PR #312, opened 2015-07-19
change nameRegEx,removing the slash[/]

lynchlee:
see jira issues USERGRID-868 .


johnament:
this should probably target master, no?

lynchlee:
Hi Johnament, thx .  I checked out a new branch from the branch "1.0.2", and do some fixes , so i targeted 1.0.2 . right ?

GERey:
Yes I agree @lynchlee and I can merge this into 1.0.2 but since most people are directed to the master branch that is where it would get the most visibility it is suggested that you point this PR to master instead.  We base our releases off of master so a PR to master would also go into future releases instead of just having a single PR to the 1.0.2 branch. Thanks!

johnament:
The 1.0.2 release already occurred.  We don't re-release.

lynchlee:
OK, got it .

michaelarusso:
@lynchlee can you close this PR?  This is a mirrored repository so you will need to do the close.  Also the branches have been renamed and the 1.0.x versions of code are now being maintained in the 1.x branch and not master.

PR #323, opened 2015-07-27
move to single map converter codebase.  

shawnfeldman:
removed core converter, only in model now. 

location fields: convert to lat lon for ES otherwise its latitude longitude, used to use lat/lon only for storage  which is not expected in api returns.  so eliminate the return conversion.


stack/core/src/main/java/org/apache/usergrid/corepersistence/util/CpEntityMapUtils.java
```diff
@@ -160,217 +101,8 @@ public static Entity fromMap(
...
-                entityMap.put(field.getName(), ((String) field.getValue()));
-
-            } else if (f instanceof LocationField) {
-                LocationField locField = (LocationField) f;
-                Map<String, Object> locMap = new HashMap<String, Object>();
-
-                // field names lat and lon trigger ElasticSearch geo location
-                locMap.put("lat", locField.getValue().getLatitude());
-                locMap.put("lon", locField.getValue().getLongitude());
-                entityMap.put( field.getName(), field.getValue());
```
shawnfeldman: serializes wrong value

PR #325, opened 2015-07-28
Fixed by upgrading the jersey version so client.create calls wouldn't…

GERey:
… create the mime files.


GERey:
https://java.net/jira/browse/JERSEY-2632 details the the error that was here https://issues.apache.org/jira/browse/USERGRID-740

PR #334, opened 2015-08-04
Usergrid 831

GERey:
https://issues.apache.org/jira/browse/USERGRID-831


snoopdave:
Shouldn't this be done for the master branch, in case we need to release 1.0.3, etc.?

GERey:
The ticket was marked for 2.1. I can point it at master as well

snoopdave:
I think we should merge it to master then from there to 2.1.

GERey:
Alright, I'll close and repoint

PR #345, opened 2015-08-11
Adds a simple 30 second flush interval task

tnine:


GERey:
Merged! Please delete your branch when you are able.

PR #353, opened 2015-08-22
Misc changes

mdunker:
1. Update to version 2.1.7 of gatling (including Scala version 2.11.7)
2. Add audit script to get all collections in all apps for an org.
3. Add a flag to control whether failed requests are printed or not.
4. Add test allowing creation of large entities. Can specify entities with m properties, each of which is a string with n kb characters.
5. Add test allowing query-based retrievals by random name.


michaelarusso:
:+1:

PR #363, opened 2015-08-28
USERGRID-918: add query parameter to limit checks for connections

mdunker:
connections=all (or no connections parameter): both connections and connecting metadata retrieved
connections=none: neither connections nor connecting metadata retrieved
connections=in: only connecting metadata retrieved
connections=out: only connections metadata retrieved

Will add tests in separate PR.


michaelarusso:
:+1:

PR #364, opened 2015-08-28
Add test delay to stale index cleanup on deletes.

michaelarusso:


michaelarusso:
@tnine this test fails intermittently on the build server.  what do you think about adding delay in here to make it more consistent?

PR #375, opened 2015-09-04
Update hasNextPage to return boolean

wooliet:
The `hasNextPage` method returned evaluation of its `_next` property,
which is either a string or null. Method documentation says it returns a
boolean value.

The result is now `_next` forced to boolean using `!!`.

### DIFF
--- a/sdks/nodejs/lib/usergrid.js
+++ b/sdks/nodejs/lib/usergrid.js
@@ -1112,7 +1112,7 @@ var AUTH_NONE = 'NONE';
     var connectee = this.getEntityId(entity);
     if (!connectee) {
       if (typeof(callback) === 'function') {
-        var error = 'Error trying to delete object - no uuid specified.';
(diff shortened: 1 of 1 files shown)


shawnfeldman:
why not just say (_next !== null && _next !== undefined) 
not sure we want to propagate !! around
also have you signed an apache icla?

wooliet:
Sorry, I have not signed anything. The readme mentions "For larger PRs" and I didn't think this would be considered a large pull request.

Using `!!` is concise and covers all falsey values, including `_next` set to an empty string. I don't know if that's possible or not with the `fetch` request on line 1760, but it seemed like the safest bet.

shawnfeldman:
can you sign and email the icla to the secretary? https://www.apache.org/licenses/icla.txt

shawnfeldman:
secretary@apache.org.

michaelarusso:
@wooliet Thanks for the contribution! I'll merge this now.

PR #380, opened 2015-09-10
USERGRID-645: fix superuser access via token

mdunker:
1. Superuser access via token now works. Some code was checking for hardcoded string "sysadmin" -- superuser role is now "service-admin".
2. Superuser token should now be granted access to all orgs and apps.
3. Locked down org config GET/PUT to superusers only.

Basic auth superuser still does not have access to orgs/apps. Org config GET/PUT requires both org access to specified org and superuser access, so token must be used for org config.


michaelarusso:
:+1:

PR #381, opened 2015-09-11
Revert "Ensure that case is being preserved when setting entity prope…

michaelarusso:
…rties."

This reverts commit 10bd033494d0cf8bd5d23db1c25665453980a5a5.


tnine:
+1, we don't want to change this in 1.0, there's a lot of internal indexing behavior that depends on it.

PR #385, opened 2015-09-16
change parsing logic to parse uuid as string

shawnfeldman:


stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsQueryVistor.java
```diff
@@ -362,8 +362,8 @@ public void visit( Equal op ) throws NoIndexException {
         final Object value = op.getLiteral().getValue();
 
         //special case so we support our '*' char with wildcard
-        if ( value instanceof String ) {
-            final String stringValue = (( String ) value).toLowerCase().trim();
+        if ( value instanceof String || value instanceof UUID ) {
+            final String stringValue = (value.toString()).toLowerCase().trim();
```
shawnfeldman: not the biggest fan of this. could be more verbose and cast string and tostring uuid

PR #386, opened 2015-09-17
Usergrid 1018

tnine:
This addresses duplicate connections that were created in previous releases. This prevents future multiple connection edges, as well as adds a task to compact them into a single connection retaining the oldest created connection (the default behavior from 1.0)


stack/core/src/main/java/org/apache/usergrid/corepersistence/service/ConnectionServiceImpl.java
```diff
@@ -117,4 +140,68 @@ public ConnectionServiceImpl( final PipelineBuilderFactory pipelineBuilderFactor
...
+                    //load edges from the source the with type specified
+                    return gm.loadEdgesFromSource( searchByEdge );
+                } )
+
+                    //now that we have a stream of edges, stream all versions
+                .flatMap( edge -> {
+
+                    logger.debug( "Found edge {}, searching for multiple versions of edge", edge );
+
+                    final SearchByEdge searchByEdge =
```
tnine: TODO: We should discuss semantics of re POSTing a connection.  If we want to retain most recent post, we'll need to change this migration as well as the write logic in Relation manager

GERey:
What would be the negatives of taking the most recent post? Isn't that what would have the most recent data?

GERey:
besides that looks good to me

PR #388, opened 2015-09-22
Changes implementation to update a connection to have the timestamp i…

tnine:
…t was last POSTed.

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/corepersistence/CpRelationManager.java
+++ b/stack/core/src/main/java/org/apache/usergrid/corepersistence/CpRelationManager.java
@@ -713,15 +713,16 @@ public class CpRelationManager implements RelationManager {
             if ( logger.isDebugEnabled() ) {
                 logger.debug( "Marking edge {} for deletion", edgeToDelete );
             }
-            return gm.markEdge( edge );
+            return gm.markEdge( edgeToDelete );
         } ).lastOrDefault( null ).doOnNext( lastEdge -> {
             //no op if we hit our default
             if(lastEdge == null){
                 return;
             }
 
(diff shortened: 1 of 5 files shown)


stack/core/src/main/java/org/apache/usergrid/corepersistence/CpRelationManager.java
```diff
@@ -697,25 +695,35 @@ public ConnectionRef createConnection( String connectionType, EntityRef connecte
...
+        gm.writeEdge( edge ).subscribe();
 
-        final SearchByEdge searchByEdge = new SimpleSearchByEdge(edge.getSourceNode(), edge.getType(), edge.getTargetNode(), Long.MAX_VALUE, SearchByEdgeType.Order.DESCENDING, Optional.absent()  );
+        indexService.queueNewEdge( applicationScope, targetEntity, edge );
 
 
-        //only take 1 and count it.  If we don't have anything, create the edge
-        final int count = gm.loadEdgeVersions( searchByEdge ).take( 1 ).count().toBlocking().last();
+        //now read all older versions of an edge, and remove them.  Finally calling delete
+        final SearchByEdge searchByEdge =
```
GERey: What are the performance implications of doing this delete every time we write a connection? Seems like every connection write be 2x the time since delete is a heavy operation. Or is there something I'm missing?

tnine: I think we'll be ok.  I agree, this may not be ideal from a performance issue.  However, I think (practically) at runtime, the most we'll be removing is 1 edge since we'll always be removing the old one after write.

GERey:
lgtm

PR #389, opened 2015-09-24
Remove buffer

shawnfeldman:


stack/corepersistence/queryindex/src/main/java/org/apache/usergrid/persistence/index/impl/EsIndexProducerImpl.java
```diff
@@ -0,0 +1,209 @@
...
+                    log.debug("adding operation {} to bulkRequestBuilder {}", batchOperation, bulkRequestBuilder);
+                    batchOperation.doOperation(client, bulkRequestBuilder);
+                }))
+                //write them
+            .doOnNext(bulkRequestBuilder -> sendRequest(bulkRequestBuilder));
+
+
+        //now that we've processed them all, ack the futures after our last batch comes through
+        final Observable<IndexOperationMessage> processedIndexOperations =
+            requests.lastOrDefault(null).flatMap(lastRequest -> {
```
tnine: I think you can get rid of this lastOrDefaultRight?  Won't it always contain the batch?

PR #396, opened 2015-10-13
remove observable

shawnfeldman:
changed observable to java 8 stream. 

added exception handling for empty sets, 1 case needs to be ignored

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/corepersistence/asyncevents/AmazonAsyncEventService.java
+++ b/stack/core/src/main/java/org/apache/usergrid/corepersistence/asyncevents/AmazonAsyncEventService.java
@@ -25,6 +25,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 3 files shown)


stack/core/src/main/java/org/apache/usergrid/corepersistence/asyncevents/AmazonAsyncEventService.java
```diff
@@ -259,71 +259,71 @@ public void ack(final List<QueueMessage> messages) {
...
+                    //collect all of the
+                    IndexOperationMessage indexOperationMessage =
+                        indexoperationObservable
+                            .collect(() -> new IndexOperationMessage(), (collector, single) -> collector.ingest(single))
+                            .toBlocking().lastOrDefault(null);
+
+                    if (validateEmptySets && (indexOperationMessage == null || indexOperationMessage.isEmpty())) {
+                        logger.error("Received empty index sequence message:({}), body:({}) ",
+                            message.getMessageId(), message.getStringBody());
+                        throw new Exception("Received empty index sequence.");
```
tnine: Won't throwing an exception here cause the entire stream to blow up, not acking messages that should work and be processed?

PR #397, opened 2015-10-15
Update the visibility timeout used when taking messages from SQS as i…

michaelarusso:
…t overrides the queue's configured/default timeout.  Also change the queue's visibility timeout property to be exposed as milliseconds for consistency.


jwest-apigee:
thumbs up.

mdunker:
:+1:

PR #399, opened 2015-10-16
Add sourceRegion to SQS event message based on queue primary region. …

michaelarusso:
… Remove cached entity manager (unused).


tnine:
LGTM

jwest-apigee:
:+1:

PR #400, opened 2015-10-19
Usergrid 1048

tnine:
Changes our ES writes in the following way.

1) Events are queued in the same region
2) Events are processed in the same region
3) Batches are then persisted as a JSON body to the map
4) Events are sent to the topic
5) Each region processes the event.


michaelarusso:
looks good

PR #411, opened 2015-10-23
Usergrid 1052

tnine:
Adds the ability to perform read-repair when reading edges that contained deleted data.


michaelarusso:
:+1:  looks good

PR #413, opened 2015-10-26
Add endpoint which lists all orgs in the system (up to 10k) and provi…

michaelarusso:
…des all apps within the org as well.

GET /management/organizations   (w/ superuser token)


tnine:
Lets keep this as a first pass.  Eventually we'll want to do paging to ensure that we don't OOM on a large response

PR #415, opened 2015-10-27
Change read repair to delete from c* and send de-index only event to all regions.

michaelarusso:
Also:
-set the default visibility timeout to 30 sec to account for systems under load 
-default # of consumers to 16
-remove schedulers.io subscription from node delete observable so we can use async scheduler higher up in code


tnine:
Solid work.

+1

PR #417, opened 2015-10-28
Convert search in ES to use term queries instead of scroll with filt…

michaelarusso:
…ers.

-Fix delete so we remove connections from Elasticsearch as well as the entity on entity delete event.


mdunker:
:+1:

PR #418, opened 2015-10-29
First and untested stab at multi-tenant migration

snoopdave:
This PR adds a new multitenant_migrate.py script that can, if used correctly, migration a Usergrid 2.0 cluster with multiple "tenants" or organizations to Usergrid 2.1 format. See the comments in the header of the file for instructions on how to run.

Associated JIRA issue: https://issues.apache.org/jira/browse/USERGRID-1068

### DIFF
--- a/stack/scripts/multitenant_migrate.py
+++ b/stack/scripts/multitenant_migrate.py
@@ -226,6 +226,7 @@ class Migrate:
                 # Init the migration system as this is the first migration done on the cluster
 
                 self.run_database_setup()
+                self.run_database_bootstrap()
 
                 migration_system_updated = self.is_migration_system_updated()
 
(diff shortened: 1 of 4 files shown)


stack/scripts/multitenant_migrate.py
```diff
@@ -0,0 +1,701 @@
...
+                self.logger.info("Finished...")
+
+                return
+
+
+            if self.init:
+
+                # Init the migration system as this is the first migration done on the cluster
+
+                self.run_database_setup()
```
michaelarusso: Should we also run self.run_database_bootstrap() here?

snoopdave: Good catch. I just added it in and tested it. I wonder how/why things worked before, I thought bootstrap was a requirement.

michaelarusso: I think if  you've already ran it manually at least once, then you're fine.  It sets up the default usergrid org, superuser acct, and the test org/account

michaelarusso:
:+1:

PR #419, opened 2015-10-29
Usergrid 1079 2x

tnine:
Cherry picks branch from USERGRID-1079-1x into this branch, also adds PUT functionality, and tests the password GET + PUT works correctly

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/persistence/CredentialsInfo.java
+++ b/stack/core/src/main/java/org/apache/usergrid/persistence/CredentialsInfo.java
@@ -20,7 +20,10 @@ package org.apache.usergrid.persistence;
 import com.fasterxml.jackson.annotation.JsonAnyGetter;
 import com.fasterxml.jackson.annotation.JsonAnySetter;
 import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.google.common.base.Preconditions;
+
 import java.io.Serializable;
+import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 import javax.xml.bind.annotation.XmlRootElement;
@@ -118,6 +121,11 @@ public class CredentialsInfo implements Comparable<CredentialsInfo>,Serializable
     }
 
 
+    public void setCreated( final Long created ) {
+        this.created = created;
+    }
+
+
     @JsonAnySetter
     public void setProperty( String key, Object value ) {
         properties.put( key, value );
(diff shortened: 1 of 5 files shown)


stack/core/src/main/java/org/apache/usergrid/persistence/CredentialsInfo.java
```diff
@@ -172,4 +180,42 @@ public int compareTo( CredentialsInfo o ) {
...
+        final boolean recoverable = ( boolean ) json.get( "recoverable");
+
+        final boolean encrypted = ( boolean ) json.get("encrypted");
+        final String hashType = ( String ) json.get( "hashType");
+        final long created = ( long ) json.get( "created");
+        final String secret = ( String ) json.get(  "secret" );
+
+        final List<String>  cryptoChain = ( List<String> ) json.get( "cryptoChain");
+
+        Preconditions.checkNotNull(created, "created is required");
```
michaelarusso: do we need to check encrypted and hashType for null values as well?

stack/rest/src/main/java/org/apache/usergrid/rest/applications/users/UserResource.java
```diff
@@ -164,6 +167,81 @@ public JSONWithPadding setUserPasswordPut( @Context UriInfo ui, Map<String, Obje
         return new JSONWithPadding( response, callback );
     }
 
+    @GET
+    @RequireSystemAccess
+    @Path("credentials")
+    public JSONWithPadding getUserCredentials(@QueryParam("callback") @DefaultValue("callback") String callback )
+            throws Exception {
+
+        logger.info( "UserResource.setUserPassword" );
```
michaelarusso: this is the GET path

stack/rest/src/main/java/org/apache/usergrid/rest/applications/users/UserResource.java
```diff
@@ -164,6 +167,81 @@ public JSONWithPadding setUserPasswordPut( @Context UriInfo ui, Map<String, Obje
...
+
+
+        final ApiResponse response = createApiResponse();
+        response.setAction( "get user password" );
```
michaelarusso: maybe let's be more explicit on what this returns as it's not the password in plaintext. 'get user password hash' or something?

stack/rest/src/test/java/org/apache/usergrid/rest/applications/collection/users/UserResourceIT.java
```diff
@@ -1103,4 +1109,107 @@ public void queryForUserUuids() throws Exception {
...
+
+
+        //get out the hash and change it so we can validate
+        final String originalSecret = ( String ) credentials.get( "secret" );
+
+
+        //here we modify the hash a little, this way we can break password validation, then re-set it to ensure we're actually updating the credentials info correctly.
+        final String borkedSecret = originalSecret.substring( 0, originalSecret.length() -1 );
+
+        credentials.put( "credentials", borkedSecret );
```
mdunker: Is this right? Or borked?

stack/core/src/main/java/org/apache/usergrid/persistence/CredentialsInfo.java
tnine: No, they're usually null

PR #423, opened 2015-11-02
Usergrid 1076

GERey:
also includes usercollection fix.


GERey:
Don't ack just yet. Still WIP

PR #427, opened 2015-11-04
Fix issue where read repair was causing empty messages to be queued.

michaelarusso:


mdunker:
code :+1: , branch name :-1:

PR #428, opened 2015-11-05
Query order of operations error

snoopdave:
Fix for https://issues.apache.org/jira/browse/USERGRID-1099

@tnine I have a simple test for bug, but I wonder: do you have any suggestions for existing tests that should be updated to more thoroughly test order of operations errors like this?


snoopdave:
The existing IntersectionTransitivePagingIT looks good, but it tests a more simple query, one that intersection two slice iterators -- so it did not tickle this bug.  

I think we are ready to merge this into 1.x. If we want more complex query tests we should open up a new JIRA for that.

tnine:
+1.  This looks ready to merge

PR #432, opened 2015-11-07
2.1 release

tnine:
Address missing invocation of unique ledger cleanup on write.  We need to ensure we clean up the ledger as we mutate unique values.


tnine:
Ignore, wrong branch :(

PR #433, opened 2015-11-07
2.1 release

tnine:
Addresses missing unique repair on write


michaelarusso:
This happens synchronous to the write commit, adding more write ops to the path when ... since it's not used anywhere now, looks we need to get it in to prevent wide rows in that CF.  However, we should spend time in master to see if we can make this an async event for deleting old versions of a unique value.

stack/corepersistence/collection/src/test/java/org/apache/usergrid/persistence/collection/EntityCollectionManagerIT.java
```diff
@@ -315,6 +315,61 @@ public void writeAndGetField() {
...
+        assertNotNull( "Version was assigned", createReturnedSecond.getVersion() );
+
+        Id idFirst = manager.getIdField( newEntity.getId().getType(), field ).toBlocking().lastOrDefault( null );
+
+        assertNull(idFirst);
+
+        Id idSecond = manager.getIdField( newEntity.getId().getType(), fieldSecond ).toBlocking().lastOrDefault( null );
+
+        assertNotNull( idSecond );
+        assertEquals( newEntity.getId(), idSecond );
```
michaelarusso: To be safe, we should assert the versions (or the versions' timestamps) as well to make sure they are different.

tnine:
I think we need to address this in 2 ways.

1) This should fire an asynchronous task.

2) Create multiple scheduler instances of the RxTaskScheduler.  1 for async queue processing. 1 for collections.

tnine:
This is done.  We need to implement read-repair on field seek to ensure that values exist and match the entity.  Unfortunately we can't guarantee consistency since repair runs after write commit and before returning to the caller.

PR #435, opened 2015-11-10
Fix filter annotations

snoopdave:
Fix annotations so that CORS and tracing filters work, and add JIRA issue for the TODO on the Metering Filter.


mdunker:
:+1: 
I'll merge, then validate the CORS fix.

PR #438, opened 2015-11-17
Removed cassandra calls to hector that surrounded every create call. …

GERey:
…Should result in less ops to cassandra but not a major cause of pain.


GERey:
https://issues.apache.org/jira/browse/USERGRID-921

PR #445, opened 2015-11-23
USERGRID-933: organization config changes

mdunker:
1) remove concept of an organization for the management app -- now the management app is treated as a special case so we don't need to have an org (and therefore a migration)
2) make 5 initial config items work
3) add query parameter separate_defaults to org config GET and PUT to display organization overrides and defaults separately in response -- default is to display org overrides overlaid over defaults
4) add query parameter items to org config GET to specify the config item(s) to be returned
5) add query parameter only_changed to org config PUT to only return information for config item(s) that were changed


mdunker:
this is old, closing

PR #446, opened 2015-11-24
Provide some extra validation of our latitude and longitude location …

michaelarusso:
…fields before adding it to the Elasticsearch index document.  This is to avoid rejected index requests due to number format exceptions.


mdunker:
:+1:

PR #447, opened 2015-12-04
Add additional functionality in REST layer for checking permissions.

michaelarusso:


snoopdave:
Looks good overall, but why do we need to disable read-repairs?

If we have to disable read repairs we should mark the tests that fail (see below) with the Ignore annotation and a comment that explains it.

Failed tests:
  ServiceRequestIT.testRepairOfSingleEntity:160 Create call should have worked
  ServiceRequestIT.testRepairOfMultipleEntitiesAndRemainingEntities:313 null

michaelarusso:
Read repair is being temporarily removed until we can test/validate further.  Good catch on the tests.  Fixing that now.

stack/rest/src/main/java/org/apache/usergrid/rest/applications/ApplicationResource.java
```diff
@@ -38,6 +38,7 @@
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.UriInfo;
 
+import org.apache.usergrid.rest.security.annotations.CheckPermissionsForPath;
```
mdunker: this is not used in ApplicationResource.java

michaelarusso: fixed.

PR #449, opened 2015-12-12
Unique index logging

GERey:


stack/tools/src/main/java/org/apache/usergrid/tools/ManagementUserAudit.java
```diff
@@ -0,0 +1,209 @@
...
+                logger.error( "Email: {} with uuid: {} doesn't exist in ENTITY_PROPERTIES.", extractedEmail, uuid );
+            }
+            else {
+                logger.info( "Email: {}  with uuid: {} exists in ENTITY_PROPERTIES for ENTITY_UNIQUE", extractedEmail,
+                        uuid );
+            }
+        }
+        else {
+            if ( cols.size() == 0 ) {
+                logger.error( "Email: {} doesn't exist in ENTITY_UNIQUE.", extractedEmail );
```
michaelarusso: Can we make this a WARN?  It's okay if it's not there, we mostly want to see errors when it's there, but has no data in ENTITY_PROPERTIES.

stack/tools/src/main/java/org/apache/usergrid/tools/ManagementUserIndexAudit.java
```diff
@@ -0,0 +1,150 @@
...
+                    logger.error( "Email: {} with uuid: {} doesn't exist in ENTITY_PROPERTIES.", extractedEmail,
+                            r.getRef().getUuid() );
+                }
+                else {
+                    logger.info( "Email: {}  with uuid: {} exists in ENTITY_PROPERTIES", extractedEmail,
+                            r.getRef().getUuid() );
+                }
+            }
+            else{
+                logger.error( "Email: {} doesn't exist in ENTITY_INDEX.",extractedEmail );
```
michaelarusso: We might want to make this a WARN as well.

PR #451, opened 2015-12-21
Add index repair to merge duplicates.

snoopdave:
Adding in Russo's new updateOrganizationUniqueIndex() to ensure that the keeper organization is the one that is in the unique name index.

### DIFF
--- a/stack/tools/src/main/java/org/apache/usergrid/tools/DuplicateOrgRepair.java
+++ b/stack/tools/src/main/java/org/apache/usergrid/tools/DuplicateOrgRepair.java
@@ -538,7 +538,7 @@ public class DuplicateOrgRepair extends ToolBase {
             // fix the org name index
             OrganizationInfo orgInfoKeeper = managementService.getOrganizationByUuid( duplicate.getId() );
             try {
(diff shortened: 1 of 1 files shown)


stack/tools/src/main/java/org/apache/usergrid/tools/DuplicateOrgRepair.java
```diff
@@ -529,10 +529,22 @@ public void call(Subscriber<? super OrgUser> subscriber) {
...
+            
+            // rename org so that it is no longer a duplicate
             OrganizationInfo orgInfo = managementService.getOrganizationByUuid( duplicate.getId() );
             orgInfo.setName( "dup_" + keeper.getId() + "_" + RandomStringUtils.randomAlphanumeric(10) );
             managementService.updateOrganization( orgInfo );
+
+            // fix the org name index
+            OrganizationInfo orgInfoKeeper = managementService.getOrganizationByUuid( duplicate.getId() );
+            try {
+                managementService.updateOrganizationUniqueIndex( orgInfo, duplicate.getId() );
```
michaelarusso: this should take in the keeper org's OrganizationInfo, and the duplicate's ID.  Looks like we're passing in the duplicates OrganizationInfo as well as the ID.

PR #453, opened 2015-12-31
USERGRID-1189: fix confusing error messages in REST layer

mdunker:


stack/rest/src/main/java/org/apache/usergrid/rest/management/users/organizations/OrganizationsResource.java
```diff
@@ -113,8 +113,8 @@ public ApiResponse newOrganizationForUserFromForm( @Context UriInfo ui, Map<Stri
         ApiResponse response = createApiResponse();
         response.setAction( "new organization for user" );
 
-        if ( organizationName == null ) {
-            throw new ManagementException( "Could not find organization for name: " + organizationName );
+        if ( organizationName == null || organizationName.isEmpty() ) {
+            throw new ManagementException( "Organization not specified." );
```
GERey: I would think we'd want to still throw back the name of the organization we haven't found in order to search through logs better? Is there a reason we wouldn't want to do that?

mdunker: This only happens if the organization name is null or an empty string. What do we have to return? :)

michaelarusso:
:+1:

PR #462, opened 2016-01-11
Usergrid 1012

GERey:
Created test to address possible issue noted here

https://issues.apache.org/jira/browse/USERGRID-1012


michaelarusso:
:+1:

PR #463, opened 2016-01-12
USERGRID-1206 UGClient does not use URLConnectionFactory.

alphamu:
User DefaultURLConnectionFactory is not URLConnectionFactory is specified.
use urlConnectionFactory.openConnection(urlAsString) instead of new URL(urlAsString).openConnection();


snoopdave:
Looks good to me. Thanks Ali!

PR #476, opened 2016-01-29
Fix dead links on documentation

kwiky:


mdunker:
:+1: Thanks for the fix!

PR #484, opened 2016-02-16
Simplify Central SSO

snoopdave:
DO NOT MERGE REVIEW PURPOSES ONLY

Instead of requiring that clients first validate tokens obtained from an external "Central SSO" system, when Usergrid is configured for using Central SSO, it should attempt to verify any unknown token itself.

This is implemented by moving the validate external token logic from the ManagementResource and into the TokenService.

https://issues.apache.org/jira/browse/USERGRID-1267


michaelarusso:
Review has been completed.  :+1:

PR #486, opened 2016-02-19
Enabled application delete without test property enabling it.

GERey:
Changed the confirm text to be "application_identifier". This needs to be set to whatever application id that you're using to refer to the application. So if /app/uuid then application_identifier is uuid but if its /app/name then you can specify it by the application name. 

Also added system default collections to be displayed and returned from the application. They are still empty but now roles isn't all alone when you have a new app :).


GERey:
Wrong branch name, reopening after.

PR #488, opened 2016-02-22
Default collections and application delete

GERey:


stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManager.java
```diff
@@ -745,7 +745,32 @@ public RelationManager getRelationManager( EntityRef entityRef ) {
     @Override
     public Set<String> getApplicationCollections() throws Exception {
 
-        return getRelationManager( getApplication() ).getCollections();
+        Set<String> existingCollections = getRelationManager( getApplication() ).getCollections();
+
+        //Handles reading custom collections.
+//        Set<String> dynamic_collections = cast( getDictionaryAsSet( getApplicationRef(), Schema.DICTIONARY_COLLECTIONS ) );
```
michaelarusso: is this left here as a TODO?

stack/rest/src/main/java/org/apache/usergrid/rest/management/organizations/applications/ApplicationResource.java
```diff
@@ -482,20 +482,27 @@ public ApiResponse executePut(  @Context UriInfo ui, String body,
...
 
-        Properties props = management.getProperties();
-
-        // for now, only works in test mode
-        String testProp = ( String ) props.get( "usergrid.test" );
-        if ( testProp == null || !Boolean.parseBoolean( testProp ) ) {
-            throw new UnsupportedRestOperationException("Test props not not functioning correctly.");
-        }
+//        Properties props = management.getProperties();
+//
```
michaelarusso: Can we remove the commented code?

stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManager.java
GERey: This was left here as a way to see how custom collections used to be queried but probably should be deleted and replaced with a jira ticket.

stack/rest/src/main/java/org/apache/usergrid/rest/management/organizations/applications/ApplicationResource.java
GERey: This one was my bad. The commented out code should be good.

PR #491, opened 2016-03-15
USERGRID-1266: check permissions at REST layer to avoid incorrect res…

mikedunker-apigee:
…ponse codes

also block user token exchange via GET /org/app/users/{user}/token?access_token={usertoken}
and log guest user into app when no token sent in for @CheckPermissionsForPath


mdunker:
IGNORE THIS ONE

PR #501, opened 2016-04-13
Added code fix for 500 errors when not using a token. Now accepts cli…

GERey:
…ent credentials as well.


snoopdave:
LGTM. Merging.

PR #504, opened 2016-04-14
Update for new Java SDK.

RobertWalsh:


sdks/java/.bash_profile
```diff
@@ -0,0 +1 @@
+export JAVA_HOME=/Library/Java/JavaVirtualMachines/jdk1.8.0_73.jdk/Contents/Home
```
GERey: Is this meant to override the current JAVA_HOME? Is there a reason we do this for the java sdk?

PR #506, opened 2016-04-15
Counter for notifications ~*-DO NOT MERGE-*~

GERey:
not ready, just to illustrate how counters would work if they did. But they don't.

### DIFF
--- a/stack/services/src/main/java/org/apache/usergrid/services/notifications/TaskManager.java
+++ b/stack/services/src/main/java/org/apache/usergrid/services/notifications/TaskManager.java
@@ -72,7 +72,7 @@ public class TaskManager {
                 }
                 successes.incrementAndGet();
             }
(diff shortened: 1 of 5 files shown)


stack/services/src/main/java/org/apache/usergrid/services/notifications/TaskManager.java
```diff
@@ -73,7 +72,9 @@ public void completed(Notifier notifier, Receipt receipt, UUID deviceUUID, Strin
                 }
                 successes.incrementAndGet();
             }
+            em.incrementAggregateCounters( notification.getUuid(),null,notification.getType(),"completed",100000L );
```
GERey: this increments the counter. by 10k or something. So change that to be sure that it isn't so high. Then do a test to make sure that increment properly increments.

Then find a way to attach these to the properties in notifications and get properly read back out. That seems mysterious to me .

michaelarusso:
Did we already add in code for this? (although commented at this point).  If so, let's close this PR.

PR #513, opened 2016-05-01
Introducing Datastax driver.

michaelarusso:


michaelarusso:
`MapSerializationImpl` and `ScopedCacheSerializationImpl` converted to use DataStax driver and CQL.

michaelarusso:
`UniqueValueSerializationStrategy` converted to use DataStax driver and CQL.  The last few commits are an example of how to work with existing DynamicComposite types from Astyanax with CQL.  Not easy, but possible.  This is temporary until we re-write the Usergrid schema into something more CQL-friendly.

PR #516, opened 2016-05-05
Service manager's init() method now retries if unable to get application

snoopdave:
And there are two configuration properties to control how often and how many times to retry.


mdunker:
👍

PR #517, opened 2016-05-06
Release class cast exception

GERey:
Fix for class cast exceptions when migrating for notifications.

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/persistence/entities/Notification.java
+++ b/stack/core/src/main/java/org/apache/usergrid/persistence/entities/Notification.java
@@ -17,17 +17,27 @@
 package org.apache.usergrid.persistence.entities;
 
 
-import javax.xml.bind.annotation.XmlRootElement;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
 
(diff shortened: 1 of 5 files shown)


michaelarusso:
Could we just keep Long and handle the casting in the notifications service where it comes in from the API (to make sure the number is put in the properties map as long instead of int):

org.apache.usergrid.services.notifications.NotificationsService#postCollection

Seems cleaner if we just handle the case there rather than making the Entity definition to a generic 'Object' and doing explicit casting in multiple places.

GERey:
Would totally agree if it wasn't for the following two issues:

1.  People already have Integers in statistics. By not accounting for Integers we force everybody who had notification statistics to not have access without them doing a re-put of all afflicted data.

2. The endpoint you mentioned is also an issue, but not the one I tested against. My main concern was with a UUID PUT with the payload. If you check my test you can see it in more detail. It would skip the NotificationService#postCollection and instead go through org.apache.usergrid.services.AbstractCollectionService#putItemById. 

The way I see it is that it is a bit ugly, but it stems from the fact that we have a field that is required to be Long but we don't explicitly enforce that on POST's or PUTs. The way we deal with this for regular entities is to useobject maps because then we can use any type based on what the user wants. Thats why I used objects above, but I can also agree that is really just a hack until we do the aforementioned checking for all of our expected types.

PR #518, opened 2016-05-09
/org/app/users/me and /users/me related fixes

snoopdave:
Ensure that /org/app/users/me always works for logged in user, regardless of roles/permissions, also that it is impossible to create an app or admin user named 'me' (via post or put)

### DIFF
--- a/stack/rest/src/main/java/org/apache/usergrid/rest/security/SecuredResourceFilterFactory.java
+++ b/stack/rest/src/main/java/org/apache/usergrid/rest/security/SecuredResourceFilterFactory.java
@@ -409,9 +409,10 @@ public class SecuredResourceFilterFactory implements DynamicFeature {
                 logger.debug( "PathPermissionsFilter.authorize" );
             }
 
-            final String PATH_MSG =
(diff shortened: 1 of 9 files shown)


stack/rest/src/main/java/org/apache/usergrid/rest/security/SecuredResourceFilterFactory.java
```diff
@@ -449,6 +450,12 @@ public void authorize( ContainerRequestContext request ) {
                 String path = request.getUriInfo().getPath().toLowerCase().replace(applicationName, "");
                 String perm =  getPermissionFromPath( em.getApplicationRef().getUuid(), operation, path );
 
+                if ( "/users/me".equals( path ) ) {
```
mdunker: Dave -- think you should allow automatic access to /users/me only for GET

snoopdave:
I think the permissions we calculate would prevent methods other than get, but adding a check here definitely makes things more clear, so I added a check for "get"

michaelarusso:
Created JIRA just FYI --> https://issues.apache.org/jira/browse/USERGRID-1289

PR #519, opened 2016-05-09
Removed tests that used events are the events are unsupported due to …

GERey:
…the problem we've encountered in our current queue system. This was decided in 2.0 and the tests were ignored in the mean time.

 I've also removed a search message test because Messages are no longer allowed to be searched on, this was also decided in 2.0.
The counters portions of the tests were kept because they are still valid.


GERey:
This PR removes events, it is the opposite to #520 since that completely fixes events. I'll leave this here for debate on which step forward should be taken.

PR #520, opened 2016-05-10
Event fixes

GERey:
Added fixes to events such that they work like they used to in 1.0.
Fixed some tests that used events since they were not translated correctly.


GERey:
This PR fixes events, it is the opposite to #519  since that completely removes events. I'll leave this here for debate on which step forward should be taken.

PR #524, opened 2016-05-16
Improvements to management app / startup logic

snoopdave:
REVIEW PURPOSES ONLY 

NOT READY FOR MERGE

Changes related to management app / startup logic 

1) Change CpEntityManagerFactory startup logic to retry when it encounters Cassandra problems when fetching the Management App.

2) Improves EntityManager cache logic (in CpEntityManagerFactory) to ensure that we do not cache EntityManagers that may be in a bad state because we cannot fetch their application data from Cassandra.  

3) Adds special EntityManager cache logic for the Management App: once we get the management app we cache it separately in case some Cassandra hiccup keeps us from fetching it we we need it later. 

4) EntityManager cache size configurable via a new property ''entity.manager.cache.size" 

https://issues.apache.org/jira/browse/USERGRID-1283

### DIFF
--- a/stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManagerFactory.java
+++ b/stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManagerFactory.java
@@ -16,11 +16,7 @@
 package org.apache.usergrid.corepersistence;
 
 
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.UUID;
+import java.util.*;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


stack/core/src/main/java/org/apache/usergrid/corepersistence/CpEntityManagerFactory.java
```diff
@@ -142,12 +141,66 @@ public CpEntityManagerFactory( final CassandraService cassandraService, final Co
...
+
+                    } else if ( managementAppEntityManager != null ) {
+
+                        // failed to fetch management app, use cached one
+                        entityManager = managementAppEntityManager;
+
+                    } else {
+
+                        // fetch failed and we have nothing cached, we must be bootstrapping
+                        logger.info("managementAppEntityManager is null, bootstrapping in progress");
```
michaelarusso: Exception is caught above when calling _getEntityManager in the case that it returns null and NPE would be thrown on getApplication().  At this point, it seems like we'd be returning a null value to the cache's load() method.  Is this intended behavior?

snoopdave:
renamed this branch to usergrid-1283-mgmt-app-init

PR #527, opened 2016-05-24
Back-out changes to ServiceManager.init() ...

snoopdave:
because we don't want retries in the HTTP request-path, at least not at this level


snoopdave:
I'm going to revert the two specific commits instead

PR #531, opened 2016-05-31
 5xx caused due to unsupported identifiers in the url param.

ayesha12:


michaelarusso:
this PR has conflicts now.  Maybe fix and open to master instead of release-2.1.1

PR #544, opened 2016-07-08
Sso2 usergrid

ayesha12:


stack/rest/src/main/java/org/apache/usergrid/rest/management/users/UsersResource.java
```diff
@@ -138,7 +138,14 @@ public ApiResponse createUser( @Context UriInfo ui, @FormParam( "username" ) Str
         ApiResponse response = createApiResponse();
         response.setAction( "create user" );
 
-        UserInfo user = management.createAdminUser( null, username, name, email, password, false, false );
+
+        UserInfo user = null;
+        if ( tokens.isExternalSSOProviderEnabled() ){
+            user = management.createAdminUser(null,username,name,email,password,true,false);
```
michaelarusso: Let's add some comments about why we're auto-activating the user in this case.

stack/rest/src/main/java/org/apache/usergrid/rest/management/organizations/users/UsersResource.java
```diff
@@ -129,11 +129,16 @@ public ApiResponse newUserForOrganizationFromForm( @Context UriInfo ui, @FormPar
         }
 
         if ( user == null ) {
-            user = management.createAdminUser( organization.getUuid(), username, name, email, password, false, false );
 
-            // A null may be returned if the user fails validation check
-            if ( user != null ) {
-                management.startAdminUserPasswordResetFlow( organization.getUuid(), user );
+            if ( tokens.isExternalSSOProviderEnabled() ){
+             user = management.createAdminUser(organization.getUuid(),username,name,email,password,true,false);
```
michaelarusso: Let's add some comments about why we're auto-activating the user in this case.

stack/services/src/main/java/org/apache/usergrid/security/tokens/cassandra/TokenServiceImpl.java
```diff
@@ -789,12 +789,17 @@ MetricsFactory getMetricsFactory() {
...
+    public boolean isExternalSSOProviderEnabled() {
         return Boolean.valueOf(properties.getProperty( USERGRID_EXTERNAL_SSO_ENABLED ));
     }
 
     private String getExternalSSOProvider(){
-        return properties.getProperty(USERGRID_EXTERNAL_PROVIDER);
+        try {
+            return properties.getProperty(USERGRID_EXTERNAL_PROVIDER);
+        }
+        catch(NullPointerException e ){
```
michaelarusso: I don't think this will throw an NPE as much as it would return ```null``` to the caller of ```getExternalSSOProvider()```

PR #547, opened 2016-07-14
Add support for validating Apigee JWTs.

michaelarusso:


michaelarusso:
1) support for apigee's SSO2
2) ability to disable public registration  ( for all non-trial pods )
3) management/me no longer generates new tokens, the token provided will be used in its response with the user data
4) superuser can now create orgs with existing admin users (before you always had to create a new admin user)
5) external/central SSO integration is generalized allow provider of Usergrid or Apigee at the moment

PR #555, opened 2016-08-02
Added cross-region, updated wait_for to use process.join() since ther…

jwest-apigee:
…e were some leftover processes in some cases.  Join() is supposed to be able to avoid this...


jwest-apigee:
This includes updates to handle a threading issue in the data migrator and other places.

PR #561, opened 2016-10-24
If claims cannot be parsed, fetch new JWT token

snoopdave:
Review purposes only. Not certain that this is the right approach.

### DIFF
--- a/stack/services/src/main/java/org/apache/usergrid/security/sso/ApigeeSSO2Provider.java
+++ b/stack/services/src/main/java/org/apache/usergrid/security/sso/ApigeeSSO2Provider.java
@@ -45,9 +45,7 @@ import java.util.Properties;
 
 import static org.apache.commons.codec.binary.Base64.decodeBase64;
 
-/**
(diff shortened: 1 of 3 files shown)


stack/services/src/main/java/org/apache/usergrid/security/sso/ApigeeSSO2Provider.java
```diff
@@ -58,9 +54,16 @@
...
     protected Client client;
     protected PublicKey publicKey;
+    protected long freshnessTime = 3000L;
+
+    public long lastPublicKeyFetch = 0L;
+
 
     public static final String USERGRID_EXTERNAL_PUBLICKEY_URL = "usergrid.external.sso.url";
 
+    public static final String USERGRID_EXTERMAL_PUBLICKEY_FRESHNESS = "usergrid.external.sso.public-key-freshness";
```
michaelarusso: Typo here on `External`

michaelarusso:
Looks good. Let's get this merged after typo fix.

PR #563, opened 2016-11-18
Configurable ElasticSearch indexing implementation

snoopdave:
Make it possible to configure either LOCAL, DISTRIBUTED or DISTRIBUTED_SNS queue; also: set Core tests to use DISTRIBUTED queue (i.e. Qakka).


michaelarusso:
👍

PR #577, opened 2017-10-17
Add Debug logging for per issues

peterj99a:


stack/corepersistence/common/src/main/java/org/apache/usergrid/persistence/core/util/DebugUtils.java
```diff
@@ -0,0 +1,73 @@
+package org.apache.usergrid.persistence.core.util;
```
michaelarusso: Needs the Apache license header.

PR #623, opened 2019-03-20
Added entity id to dictionary object in export to establish relations…

bursech:
Added entity id to dictionary object in export to establish relationship between entity and dictionary.


stack/tools/src/main/java/org/apache/usergrid/tools/Export.java
```diff
@@ -639,6 +639,7 @@ private void fetchDictionaries(File collectionDir, String collectionName, final
 							}
 							
 							jgDictionaries.writeStartObject();
+							jgDictionaries.writeObjectField("Related Entity", et.getUuid());
```
keyurkarnik: Change the tag from Related Entity to Entity since there is no other entity present, unless there is a specific reason to call it Related Entity

PR #624, opened 2019-03-22
Added entity id and type to dictionary object in export.

bursech:
This will establish relationship between entity and dictionary.


keyurkarnik:
Add Jira link

PR #626, opened 2019-05-31
USERGRID-1348 : Fix failure for CollectionDeleteTest

keyurkarnik:
The test for CollectionDelete is failing on machines with more
processing power because the timestamp for the last to-be-deleted entry
and the next one ends up being the same. Added a sleep and also setup
non-boundary test scenarios and more logging.
Updated the Jenkins build to increase timeouts

### DIFF
--- a/stack/Jenkinsfile
+++ b/stack/Jenkinsfile
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 
+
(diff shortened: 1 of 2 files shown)


keyurkarnik:
https://issues.apache.org/jira/browse/USERGRID-1348

